PR #32, opened 2022-09-13
Feature interp

### LINKED ISSUE #31: Improve SED interpolation in training script -> Add new hyperparameters for automatic interp
@CentofantiEze be careful with the interpolation of the SEDs that require reweighting of some values.

CentofantiEze:
This branch introduces the SED interpolation feature. Latest modifications allows using directly the interpolation while optimising the model instead of having to create outside an interpolated SED dataset. 

It adds new click options:
- `interp_pts_per_bin` int in {0,1,2,3}. Number of points per bin to add during the interpolation process.
- `extrapolate` bool. Whether extrapolation is performed or not.  
- `SED_interp_kind` str in {'linear', 'cubic'}. Type of interpolation for the SED. 
- `n_bins_gt` int. Number of bins used for the ground truth poly PSF generation.


CentofantiEze:
Somehow the feature_multi_cycle branch merged to this one.

Issue #40, opened 2023-01-05
Main Branch: KeyError: 'sed_sigma'


jeipollack:
Hi, 
  I attempted to run the `main` branch wf-psf code on the Apple M1 chip and encountered the `KeyError: 'sed_sigma'`.  Below is the traceback.
I do not encounter this error when running the `feature-multi-cycle` branch on the M1 chip, which seems to run to completion without any errors.
Comparing the two branches, i.e.  `main` and `feature-multi-cycle`, I see this keyword does not exist in the feature branch. 
But, this feature branch is several commits behind `main` so I am not sure whether this `sed_sigma` keyword should  exist.
  
`
Traceback (most recent call last):
  File "/Users/jenniferpollack/Projects/wf-psf/long-runs/train_eval_plot_script_click.py", line 376, in <module>
    main()
  File "/Users/jenniferpollack/miniconda/envs/wf-psf/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jenniferpollack/miniconda/envs/wf-psf/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/jenniferpollack/miniconda/envs/wf-psf/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jenniferpollack/miniconda/envs/wf-psf/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/jenniferpollack/Projects/wf-psf/long-runs/train_eval_plot_script_click.py", line 365, in main
    wf.script_utils.train_model(**args)
  File "/Users/jenniferpollack/miniconda/envs/wf-psf/lib/python3.9/site-packages/wf_psf/script_utils.py", line 125, in train_model
    SED_sigma=args['sed_sigma']
KeyError: 'sed_sigma'`

jeipollack:
I do see that the SED_sigma param is listed in the [Readme](https://github.com/CosmoStat/wf-psf/tree/main/long-runs#wavediff-new-features) as a new feature.

jeipollack:
I get the same error for the `feature-interp` branch which was one the latest PRs to `main` 

[...]

PR #52, opened 2023-01-24
Fixed missing/type args errors #40

jeipollack:
As reported in Issue #40 there was a missing arg with the last PR to the `main` branch.  Tobias confirmed an error occurred with the last PR.

```
click.option(
     "--SED_sigma",
     default=0,
     type=float,
     help="Standard deviation of the multiplicative SED Gaussian noise.")
 # Feature: project parameters

```
After running, I then encountered a few type errors, where `str` was set as default when numeric type was expected.  So, I did a check of the previous commits to see the original ranges and updated the default for the args in question.

```
@click.option(
    "--l_rate_param",
    nargs=2,
    default=[1e-2, 1e-2],
    type=float,
    help="Learning rates for the parametric parts."
)
@click.option(
    "--l_rate_non_param",
    nargs=2,
    default=[1e-1, 1e-1],
    type=float,
    help="Learning rates for the non-parametric parts."
)
@click.option(
    "--n_epochs_param",
    nargs=2,
    default=[20, 20],
    type=int,
    help="Number of training epochs of the parametric parts."
)
@click.option(
    "--n_epochs_non_param",
    nargs=2,
    default=[100, 120],
    type=int,
    help="Number of training epochs of the non-parametric parts."`
```
After these changes, I re-ran the `wf-psf` code and encountered no further training errors.

### DIFF
--- a/long-runs/train_eval_plot_script_click.py
+++ b/long-runs/train_eval_plot_script_click.py
@@ -11,11 +11,11 @@ import click
 
 
 @click.command()
-## Script options
+# Script options
 @click.option("--train_opt", default=True, type=bool, help="Train the model.")
 @click.option("--eval_opt", default=True, type=bool, help="Evaluate the model.")
 @click.option("--plot_opt", default=True, type=bool, help="Plot the model results")
-## Training options
+# Training options
 # Model definition
 @click.option(
     "--model",
@@ -61,9 +61,9 @@ import click
     help="Path to save model checkpoints during training."
 )
 @click.option(
-    "--plots_folder", 
-    default="plots/", 
-    type=str, 
+    "--plots_folder",
+    default="plots/",
+    type=str,
     help="Folder name to save the generated plots."
 )
 # Input dataset paths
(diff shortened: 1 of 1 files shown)


tobias-liaudat:
Ah @jeipollack 

I have to check, but I think the `l_rate_param`, `l_rate_non_param`, `n_epochs_param`, `n_epochs_non_param` parameters were changed to str on purpose when the functionality of the multi-cycle training was introduced.

The problem was that the click package could not naturally handle various arguments, which depended on the number of training cycles. Then, a string would do the work as it can be parsed into different arguments.

Although this change is not compatible with older job scripts where the arguments are fixed at two floats.

jeipollack:
Hi @tobias-liaudat , okay. In any case, the refactored version of the code will abandon the `Click` package to parse arguments. Instead the training params will be loaded with a `yaml` file reader.

tobias-liaudat:
@jeipollack oh good to know that! Click can be a bit annoying sometimes, it's better what you propose :)

Issue #43, opened 2023-01-20
Add `config` folder to store `yaml` files


jeipollack:

jeipollack:
Done in the branch `feature_refactor_ui`.

jeipollack:
task completed.

Issue #50, opened 2023-01-20
Implement UI with yaml config files containing program arguments to run at execution


jeipollack:

jeipollack:
A yaml config file containing the model training parameters has been set up with a yaml file reader.  Further development under way.

PR #53, opened 2023-02-09
Feature refactor UI

### LINKED ISSUE #51: Add method to automatically create directory structure for `wf-outputs` if files do not exist

jeipollack:
This PR features many changes which I subdivide into major and minor categories:

**Major**
- Added a `run.py` module to launch at execution which parses the command line arg specifying the name of the configuration file: `configs.yaml`
- Added a `.env` script storing the environment variables required to be set at execution with the `dotenv` package method `load_dotenv`
- Provided an `.env.example` in the repo which the user should update with the appropriate paths to the directories `wf-psf` repo, the directory for training and test datasets, and the location of the output directory closes #51 
- Added a `io.py` module with methods to automatically make the `wf-outputs` file structure defined in [long-runs instructions](https://github.com/CosmoStat/wf-psf/tree/feature_refactor_ui/long-runs#output-folders)
- New UI interface applied for setting training parameters resolves issues #43 and #50 with a main `configs.yaml` file which should store individual `config` files: `training_config.yaml`, `eval_config.yaml` , `plotting_config.yaml` and a `logging.conf`.  Only the `training_config.yaml` and `logging.conf` have been written and used so far.
- Created the `training` package which should contain `train.py` module with a class `TrainingParamsHandler` for extracting the different parameter subsets for training, psf model, and datasets, and a `train` method which should do the training once the psf model, training and test datasets have been specified (TO DO)
- Created a `psf_model` package containing for now a `psf_model.py` module with a decorator `register_psf_class` wrapping each PSF model class.  At runtime, a dictionary called `PSF_CLASSES` will get loaded with all of the psf_model classes. The `dict` key is determined according to the class attribute `ids` which stores the short model name of the class specified in the `training_config.yaml`.
- Moved zernikes methods from the `utils.py` module to a `zernikes.py` module 

**Minor Changes**
- Reduced `script_utils.py` to only train the `poly` model.  Have not tested it. 
- Corrected minor typos in commented versions 
- Created a `metrics` package and moved the `metrics.py` module there.
- Updated `wf_psf/__init__.py` with the `train`, `psf_models` and `metrics` packages


jeipollack:
oops. I just remembered that I need to write a method to verify that the config parameters are correct.

sfarrens:
Thanks @jeipollack, I will start reviewing this tomorrow. :)

jeipollack:
Thanks, @sfarrens. I am reviewing myself, too! I see that I need to tidy up the `io.py` module. And, since the construction of the output file directory is automatic I was thinking to convert it to internal commands. If you have any thoughts on that, let me know. Thanks!

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
+"""
+:file: wf_psf/io.py
+
+:date: 18/01/23
+:author: jpollack
+
+"""
```
sfarrens: We should decide on a documentation standard so that we can keep everything consistent going forward. My preference is [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) and ensuring that everything passes [pydocstyle](https://www.pydocstyle.org/en/stable/) tests.

In summary, for modules, it would be something along the lines of:

```python
"""MODULE NAME.

A short summary of what the modules contains.

:Author: Author's name

"""
```

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pathlib
+import os
+from dotenv import load_dotenv
+
+load_dotenv("./.env")
+_workdir = os.getenv('WORKDIR')
+_wf_outputs = "wf-outputs"
+_chkp = "chkp"
```
sfarrens: What does `chkp` stand for? Perhaps you can user a richer variable name.

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+_log_files = "log-files"
+_metrics = "metrics"
+_optim_hist = "optim-hist"
+_plots = "plots"
+
+
+def make_wfpsf_file_struct():
+    """A function to launch commands
+    to set up wf-outputs file structure.
+    """
```
sfarrens: Again, we should set a standard for the function docstrings. My personal preference for functions is shown below. In this particular case, it could be useful to have a "See Also" section for each of the called functions.

```python
"""Function name in words.

A more detailed description of what the function does.

Parameters
----------
arg1 : dtype
    Argument description

Returns
-------
dtype
    Returned object description

Raises
------
Exception type
    Reason for exception

See Also
--------
Function/method : reason for connection

Examples
--------
Testable usage example(s)

Notes
-----
Additional information, equations, paper citations, etc.

"""
```

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+
+def _make_wf_outputs_dir():
+    """
+    A function to make wf-outputs dir.
+    """
+    pathlib.Path(os.path.join(_workdir, _wf_outputs)
+                 ).mkdir(exist_ok=True)
+
+
+def _make_chkp_dir():
```
sfarrens: Missing docstring.

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+    pathlib.Path(os.path.join(_workdir, _wf_outputs,
+                              _chkp)).mkdir(exist_ok=True)
```
sfarrens: I would propose we use [Black](https://black.readthedocs.io/en/stable/) to manage the Python style formatting. This can integrated into various text editors (such as [VS Code](https://code.visualstudio.com/)) or run manually in the terminal.

In any case, for long lines it is better format them as follows.
```python
my_long_list = [
    item1,
    item2,
    ...,
]

my_long_function_call(
    arg1,
    arg2,
    ...,
)

my_super_long_string = (
    "this goes on for quite a bit ... "
    + "... and keeps going ... "
    + "... and so on."
)

```

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _make_optimhist_dir():
+    pathlib.Path(os.path.join(_workdir, _wf_outputs,
+                 _optim_hist)).mkdir(exist_ok=True)
+
+
+def _make_plots_dir():
+    pathlib.Path(os.path.join(_workdir, _wf_outputs,
+                              _plots)).mkdir(exist_ok=True)
```
sfarrens: This looks like the same function repeated over and over. If you are simply passing a different argument each time I think this can be greatly simplified.

e.g.
```python
def _make_dir(dir_name):
    pathlib.Path(os.path.join(_workdir, _wf_outputs, dir_name)).mkdir(exist_ok=True)

_make_dir(_chkp)
```

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+from dotenv import load_dotenv
+
+load_dotenv("./.env")
+_workdir = os.getenv('WORKDIR')
+_wf_outputs = "wf-outputs"
+_chkp = "chkp"
+_log_files = "log-files"
+_metrics = "metrics"
+_optim_hist = "optim-hist"
+_plots = "plots"
```
sfarrens: Since most of these variables are only used once, you could probably just make a tuple of strings and pass and loop through them (see comment below about the function redundancy).

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+                 _optim_hist)).mkdir(exist_ok=True)
+
+
+def _make_plots_dir():
+    pathlib.Path(os.path.join(_workdir, _wf_outputs,
+                              _plots)).mkdir(exist_ok=True)
+
+
+def _get_log_save_file():
+    pass
```
sfarrens: Is this for something that you are planning to implement later?

wf_psf/io.py
```diff
@@ -0,0 +1,83 @@
...
+:author: jpollack
+
+"""
+
+import pathlib
+import os
+from dotenv import load_dotenv
+
+load_dotenv("./.env")
+_workdir = os.getenv('WORKDIR')
```
sfarrens: I think you need to explain to me why we need environment variables to be defined at all?

long-runs/train_eval_plot_script_click.py
```diff
@@ -5,17 +5,18 @@
 
 import wf_psf as wf
 import click
+from wf_psf.io import make_wfpsf_file_struct
```
sfarrens, line 8: Is the idea here to keep the existing scripts in working order until you have built a more general UI?

sfarrens (changes requested):
First batch of comments. 🙂

wf_psf/io.py
jeipollack: `chkp` stands for checkpoint the original naming scheme defined by Tobias.  I could rename to simply to `checkpoint` it's not very long.

wf_psf/io.py
jeipollack: Hi @sfarrens, many thanks for these useful suggestions. I began the modifications of the modules adopting the `numpy` docstyle.  I've committed and pushed my changes for `io.py` module.  For the other modules, it's made me rethink and change some of the structure.

Currently, I am working on the `psf_models.py` module that's a combo of my new additions and old additions by Tobias.  This module contains a specific psf model, which should go into its own module as with the other PSF models.  
I also started modifying the `training.py` module in the `training` package.  There also more steps to carry out here.

I want to check if it's okay with you if I carry on the development of these two packages, i.e. PSF models and Training, before resubmitting this PR. Or, do you prefer I focus on just the specific items raised in this PR concerning UI?

wf_psf/io.py
jeipollack: Thanks for this tip.  I pip installed `Black` and installed the Black extension into VSCode, as well as updated my `settings.json` file

```
{
    "python.analysis.extraPaths": [
        "./wf_psf/training"
    ],
    "[python]": {
        "editor.defaultFormatter": null,
        "editor.insertSpaces": true,
        "editor.tabSize": 4,
        "editor.formatOnSave": true
    }
}
```

But, it doesn't automatically update upon saving.  I have to run Black manually from the command-line.  And, note I am not quite sure why I have listed `python.analysis.extraPaths` in the `settings.json`. I don't recall adding it there. :D I have to see what that does.

wf_psf/io.py
jeipollack: done!

wf_psf/io.py
jeipollack: done! thanks for the reminder of the DRY principle. :-)

wf_psf/io.py
jeipollack: I removed this function.

wf_psf/io.py
jeipollack: Yes, it was added a reminder to match the original code.  But, I decided to remove it as the log file is created at the start of the program.

long-runs/train_eval_plot_script_click.py
jeipollack: I forgot about this change. I think this particular method is not needed, so I removed it.  But, I wasn't planning to run the existing script in this branch as I trimmed it down. Now I am debating whether that is a bad idea.  I guess if I need to debug, I can bring back some of the older parts if I trimmed too much.

wf_psf/io.py
jeipollack: This `load_dotenv` is temporary.  Actually, I had initially coded this variable to be passed in as a command-line argument, but I changed to specifying environment variables so the user doesn't have to specify it for this specific program. I was thinking these variables would also be defined when running on different systems and in the Euclid interface scripts.

wf_psf/io.py
```diff
@@ -0,0 +1,131 @@
...
+def _setup_dirs():
+    """Setup Directories.
+
+    A function to setup the output
+    directories.
+
+    Parameters
+    ----------
+    list_of_dirs: tuple
+        Lists of output directories
```
sfarrens: This is not technically a parameter because it is not passed as an argument to `_setup_dirs`

wf_psf/io.py
```diff
@@ -0,0 +1,131 @@
...
+
+    A function to set up
+    logging.
+
+    Parameters
+    ----------
+    repodir: str
+        Repo directory
+    logfile: str
+        Log filename with absolute path
```
sfarrens: Same point here regarding parameters.

wf_psf/io.py
```diff
@@ -0,0 +1,131 @@
...
+
+load_dotenv("./.env")
+
+_workdir = os.getenv("WORKDIR")
+_wf_outputs = "wf-outputs"
+_checkpoint = "checkpoint"
+_log_files = "log-files"
+_metrics = "metrics"
+_optim_hist = "optim-hist"
+_plots = "plots"
```
sfarrens: These three variables only appear to be called once. Is there a reason they need to be defined at the module level?

wf_psf/io.py
```diff
@@ -0,0 +1,131 @@
+"""IO
```
sfarrens:
```suggestion
"""IO.
```

wf_psf/io.py
sfarrens: We can discuss this in our next meeting.

wf_psf/io.py
sfarrens: OK, my question is more basic. Why does the user need to specify a `WORKDIR` environment variable, rather than e.g. specifying a path in the config file?

wf_psf/io.py
jeipollack: Thanks, I was waiting for this comment bc I wasn't sure. :-)

wf_psf/io.py
jeipollack: same as above.

wf_psf/io.py
jeipollack: These are the equivalent of [L36 ](https://github.com/CosmoStat/wf-psf/blob/899a6e36502b8ddc7af99dce04c30019c653bcf8/wf_psf/script_utils.py#L36) in `script_utils.py` (although they are paths not files) and are used in the training, evaluating, and plotting methods.  I call them once in the module, but I was planning to call them again when creating and writing to file in their respective directories.

wf_psf/io.py
jeipollack: Thanks, I corrected in my local version, but didn't push. :-)

wf_psf/io.py
jeipollack: Initially,  I assumed the repository was in the path of the `WORKDIR`, because that is the case for me.  But, I decided it doesn't have to be if the user would prefer it be in a different location.  It could go into the config file. Is that what you prefer? :-)

sfarrens: My view is that things like environment variables are not as easily controlled by us (i.e. the developers). So, if someone has an issue e.g. because they manually set `WORKDIR` to something weird, it is hard for us to do anything about it. Of course the same could happen with a config file, but it is easier for us to identify the source of the issue.

wf_psf/io.py
sfarrens: OK, I think we can look at this during our next meeting also.

wf_psf/io.py
jeipollack: yes, that's a good point

wf_psf/io.py
jeipollack: I created a FileIOHandler class and moved these guys to attributes of this class.

wf_psf/io.py
jeipollack: I renamed this variable to `outputdir` and it's now a command-line arg / instance variable for the `class FileIOHandler`

wf_psf/io.py
```diff
@@ -0,0 +1,154 @@
...
+        -------
+        str
+            Absolute path to optimizer directory
+
+        """
+        return os.path.join(self._output_path, self._wf_outputs, self._optimizer)
+
+
+if __name__ == "__main__":
+    _setup_dirs()
```
sfarrens: Remove

wf_psf/io.py
```diff
@@ -0,0 +1,154 @@
...
+
+"""
+
+import pathlib
+import os
+from dotenv import load_dotenv
+import logging
+from datetime import datetime
+
+load_dotenv("./.env")
```
sfarrens: Remove

wf_psf/io.py
```diff
@@ -0,0 +1,154 @@
...
+            self._make_dir(dir)
+
+    def _setup_logging(self):
+        """Setup Logger.
+
+        A function to set up
+        logging.
+
+        """
+        repodir = os.getenv("REPODIR")
```
sfarrens: Change to config option

wf_psf/io.py
```diff
@@ -0,0 +1,154 @@
...
+A module which defines methods to
+manage wf-psf inputs and outputs.
+
+:Author: Jennifer Pollack <jennifer.pollack@cea.fr>
+
+"""
+
+import pathlib
+import os
+from dotenv import load_dotenv
```
sfarrens: Remove

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
+"""WF_PSF Run
```
sfarrens:
```suggestion
"""WF_PSF Run.
```

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import logging.config
+import logging
+from wf_psf.training import train
+
+# load .env variables
+load_dotenv("./.env")
+
+# set repo directory
+repodir = os.getenv("REPODIR")
```
sfarrens: Remove/update

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
...
+file_handler = FileIOHandler(args.outputdir)
+file_handler.setup_outputs()
+
+logger = logging.getLogger("wavediff")
+
+logger.info("#")
+logger.info("# Entering wavediff mainMethod()")
+logger.info("#")
+
+configs = read_stream(os.path.join(repodir, args.conffile))
```
sfarrens: Update

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
...
+
+for conf in configs:
+    if hasattr(conf, "training_conf"):
+        # load training_conf
+        training_params = read_conf(os.path.join(repodir, conf.training_conf))
```
sfarrens: Update

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
...
+"""WF_PSF Run
+
+This module setups the run of the WF_PSF pipeline.
+
+:Author: Jennifer Pollack <jennifer.pollack@cea.fr>
+
+"""
+import argparse
+from wf_psf.read_config import read_stream, read_conf
+from dotenv import load_dotenv
```
sfarrens: Remove

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
+"""Train
```
sfarrens:
```suggestion
"""Train.
```

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+
+    psf_model = psf_models.get_psf_model(
+        training_handler.model_name,
+        training_handler.model_params,
+        training_handler.training_hparams,
+    )
+
+    logger.info(f"PSF Model class: `{training_handler.model_name}` initialized...")
+
+if __name__ == "__main__":
```
sfarrens: Remove

sfarrens (changes requested):

wf_psf/run.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
sfarrens: You should decide if you want this to be a script (at least for now) or not.

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+    - training_params: Recursive Namespace
+        Recursive namespace object containing training input params
+    - id_name: str
+        ID name
+    - run_id_name: str
+        Run ID name
+    - output_dirs: FileIOHandler
+        FileIOHandler instance
+    - saving_optim_hist: dict
+        Dictionary storing optimiser parameters
```
sfarrens: Check, which parameters are actually being passed to `__init__`. Also no need for the `-` before each parameter name.

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+
+    """
+
+    def __init__(
+        self, training_params, output_dirs, id_name="-coherent_euclid_200stars"
+    ):
+        self.training_params = training_params.training
```
sfarrens: Maybe it would be easier to pass the `training_params.training` object directly to this class?

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+
+def get_gpu_info():
+    """Get GPU Information.
+
+    A function to return GPU
+    device name.
+
+    Returns
+    -------
+    device_name
```
sfarrens: This should be a valid data type.

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def train(training_params, output_dirs):
+    """Train.
+
+    A function to train the psf model.
+
+    Parameters
+    ----------
+    training_params: Recursive Namespace
```
sfarrens: I don't think this is a valid data type.

wf_psf/training/train.py
```diff
@@ -0,0 +1,138 @@
...
+    psf_model = psf_models.get_psf_model(
+        training_handler.model_name,
+        training_handler.model_params,
+        training_handler.training_hparams,
+    )
+
+    logger.info(f"PSF Model class: `{training_handler.model_name}` initialized...")
+
+if __name__ == "__main__":
+    workdir = os.getenv("HOME")
```
sfarrens: Remove

sfarrens (changes requested):

wf_psf/io.py
jeipollack: done.

wf_psf/io.py
jeipollack: done.

wf_psf/io.py
jeipollack: changed to command-line arg

wf_psf/io.py
jeipollack: done.

wf_psf/run.py
jeipollack: rm. it is a script for now.

wf_psf/run.py
jeipollack: Done

wf_psf/run.py
jeipollack: Done.

wf_psf/run.py
jeipollack: Done.

jeipollack:
Hi @sfarrens , I pushed doc string style changes for `read_config.py`.

wf_psf/training/train.py
jeipollack: yes, that was more a  note for myself since some of them are Tobias-defined variables :-)

jeipollack: I will rename some of the variables

wf_psf/training/train.py
jeipollack: Done.

wf_psf/training/train.py
jeipollack: corrected

wf_psf/training/train.py
jeipollack: I listed it as a `type` and noted below that is a Recursive Namespace object

wf_psf/training/train.py
jeipollack: done

jeipollack:
Hi @sfarrens,  just want to send you a FYI that I did a few more changes in this branch more related to restructuring the package, but I did not push them.  Instead I decided to checkout a new branch to continue development there until this PR can be merged to the `dummy_main` branch.

sfarrens (approved):

Issue #46, opened 2023-01-20
From `wf_psf/script_utils.py`, move training, psf modelling, data manipulation, evaluation and plotting to modules in their own packages, if necessary


jeipollack:

jeipollack:
In progress.
Created a `training` python package containing a `training.py` module that will be used to gather the input model `training params` via a Class constructor with methods to execute the training.

Working on a `zernikes.py` module and deciding on how to represent the other data products for training and testing.

jeipollack:
I completed this task in the feature branch `Refactor-psf-models`. will submit a PR shortly

Issue #44, opened 2023-01-20
Add `utils` folder with module for handling io


jeipollack:

jeipollack:
I guess I can close this issue to mark it as complete.

PR #55, opened 2023-03-16
Refactor psf models

jeipollack:
Major Changes in this PR:

- Reorganization of most of the modules to associated packages (`script_utils.py` will be made deleted in the future) associated to issues: #44 and #46 
- Added modules for `parametric` and `semi parametric` PSF models
- Added `training_preprocessing.py` module with handlers for Training and Test data

Modules to be reviewed:

- `wf_psf/data/training_preprocessing.py`
- `wf_psf/training/train.py`


wf_psf/data/training_preprocessing.py
```diff
@@ -0,0 +1,114 @@
...
+    Parameters
+    ----------
+    filename: str
+        Name of file
+    allow_pickle_flag: bool
+        Boolean flag to set when loading numpy files
+
+    """
+    dataset = np.load(filename, allow_pickle=allow_pickle_flag)[()]
+    return dataset
```
sfarrens: I think you make this one line since you don't really need the `dataset` object.

wf_psf/data/training_preprocessing.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class TrainingDataHandler:
+    """Training Data Handler.
+
+    maybe rename to TensorFlow Handler
+    or TrainingDataSetup
```
sfarrens: Is this something you want to keep in the docstring? If this is not something you plan to implement in this PR then perhaps it would be better as a developer note (i.e. `# Maybe rename ...`) or a `Todo` section.

wf_psf/data/training_preprocessing.py
jeipollack: I decided to load the file explicitly during instantiation, so I removed the function.

wf_psf/data/training_preprocessing.py
jeipollack: I forgot about this note. I removed it.

jeipollack:
lots of commits bc I forgot I need to use my personal access token to `git push` from JZ.

wf_psf/data/training_preprocessing.py
```diff
@@ -0,0 +1,93 @@
...
+            for _sed in self.train_dataset["SEDs"]
+        ]
+        self.sed_data = tf.convert_to_tensor(self.sed_data, dtype=tf.float32)
+        self.sed_data = tf.transpose(self.sed_data, perm=[0, 2, 1])
+
+
+class TestDataHandler:
+    """Test Data Handler.
+
+      A class to handle test data for model validation.
```
sfarrens:
```suggestion
    A class to handle test data for model validation.
```

wf_psf/training/train.py
```diff
@@ -12,11 +12,14 @@
...
 import tensorflow as tf
 import tensorflow_addons as tfa
-from wf_psf.read_config import read_conf
+from wf_psf.utils.read_config import read_conf
 import os
 import logging
-import wf_psf.io as io
-from wf_psf.psf_models import psf_models
+import wf_psf.utils.io as io
+from wf_psf.psf_models import *
```
sfarrens: It's not good practice to use `import *` as it creates ambiguity as to where a given object (function, class, etc.) has come from. What I would propose instead is:

a. if there are only a few objects needed, to import them explicitly,
b. if there many objects needed, to import `psf_models` as an alias, e.g. `import wf_psf.psf_models as pm`

sfarrens (commented):
Hi @jeipollack things are looking good. I opened a couple of minor threads. Some points to consider for `train.py`:

- The class method `train` is very long. I feel like a lot of these steps could be broken up into smaller "private" class methods.
- There are lots of `print` statements. Perhaps you could include a `verbose` option (which is `True` by default) in the UI and add `if verbose:` (or something to that effect) to these.

wf_psf/training/train.py
jeipollack: this is only importing these modules:

__all__ = ['psf_models', 'psf_model_parametric', 'psf_model_semiparametric']

Do you still prefer I change?

jeipollack:
Hi @sfarrens , I am ready for you to complete your review of `train.py` and the rest of this PR. The refactoring resulted in changes to `configs/training_config.yaml` and additional attributes to `psf_models/psf_model_semiparametric.py`.

jeipollack:
PR Msg to test GH -> Slack.

jeipollack:
ah I won't receive a notification bc I sent it.

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+        Recursive Namespace object containing training hyperparameters for this PSF model class
+    coeff_mat: Tensor or None
+        Initialization of the coefficient matrix defining the parametric psf field model
+
+    """
+
+    ids = ("poly",)
+
+    def __init__(self, model_params, training_params, coeff_mat=None):
+        super(TF_SemiParam_field, self).__init__()
```
sfarrens: This should work and is a bit cleaner, in particular if you decided to change the class name. 

```suggestion
        super().__init__()
```

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+    def set_output_Q(self, output_Q, output_dim=None):
+        """Set the value of the output_Q parameter.
+
+        Useful for generating/predicting PSFs at a different sampling wrt the
+        observation sampling.
+
+        Jennifer: this function doesn't make sense to me.
+        could be broken up into individual methods (i.e. getter/setter for output_Q
+        and output_dim)
+        and method to reinitialize PSF batch poly generator
```
sfarrens: I think this should just be a developer comment (i.e. starting with `#`) rather than part of the docstring if you wanted included in the PR.

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+            output_Q=self.output_Q,
+            output_dim=self.output_dim,
+        )
+
+    def predict_mono_psfs(self, input_positions, lambda_obs, phase_N):
+        """Predict a set of monochromatic PSF at desired positions.
+
+        Parameters
+        ----------
+        input_positions:
```
sfarrens: Missing data type

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+            Tensor(batch_dim x 2)
+
```
sfarrens:
```suggestion
```

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+        lambda_obs: float
+            Observed wavelength in um.
+
```
sfarrens, line 207:
```suggestion
```

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+                            inter_res_v2[j, :, :],
+                            n_pix_zernike,
+                        )
+                        for j in range(self.tf_poly_Z_field.coeff_mat.shape[1])
+                    ]
+                    for i in range(self.n_zernikes)
+                ]
+            ),
+            dtype=tf.float32,
+        )
```
sfarrens, line 297: This looks a bit messy to me. 

It doesn't need to be changed for this PR, but perhaps something we should come back to. Maybe you can add something like `# TODO: Clean up` (or something along those lines).

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+            [
+                [
+                    PI_zernikes(tf_zernike_cube[j, :, :], S_tilde_slice, n_pix_zernike)
+                    for j in range(self.n_zernikes)
+                ]
+                for S_tilde_slice in S_tilde[
+                    : self.tf_poly_Z_field.coeff_mat.shape[1], :, :
+                ]
+            ]
+        )
```
sfarrens, line 320: Same point here.

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+            [(0, S_tilde.shape[0] - beta_tilde_inner.shape[0]), (0, 0)],
+            mode="constant",
+        )
+
+        # Unmix beta tilde with the inverse of alpha
+        beta = tf.constant(
+            np.linalg.inv(self.tf_np_poly_opd.alpha_mat) @ beta_tilde, dtype=tf.float32
+        )
+        # Get the projection for the unmixed features
+
```
sfarrens, line 335: Does this comment refer to the `S_mat_projected` object? It looks a bit lost.

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+class TF_SemiParam_field(tf.keras.Model):
+    """PSF field forward model.
+
+    Semi parametric model based on the Zernike polynomial basis.
+
+    Parameters
+    ----------
+    model_params: type
+        Recursive Namespace object containing parameters for this PSF model class
+    training_params: type
```
sfarrens: What is data type `type`?

wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -0,0 +1,381 @@
...
+from wf_psf.psf_models.tf_layers import (
+    TF_poly_Z_field,
+    TF_zernike_OPD,
+    TF_batch_poly_PSF,
+)
+from wf_psf.psf_models.tf_layers import (
+    TF_NP_poly_OPD,
+    TF_batch_mono_PSF,
+    TF_physical_layer,
+)
```
sfarrens: Since most of these objects are only called once, it would probably be a lot leaner to do something like:

```python
from wf_psf.psf_models import psf_models as psfm
from wf_psf.psf_models import tf_layers as tfl
```

and then later something like e.g.:

```python
self.tf_poly_Z_field = tfl.TF_poly_Z_field(...)
```

etc.

wf_psf/training/train.py
sfarrens: Yes, I would argue we should never see `import *` in any part of the code.

wf_psf/training/train.py
```diff
@@ -100,6 +103,94 @@ def training_hparams(self):
...
 
+    @property
+    def multi_cycle_params(self):
+        """Training Multi Cycle Parameters.
+
+        Set training multi cycle parameters
+
+        Returns
+        -------
+        multi_cycle_params: type
```
sfarrens: Data type `type` again?

sfarrens (changes requested):
Hi @jeipollack here is another round of comments. Overall things are looking good, but there are still some things that could be better abstracted (this can be looked at in a future PR). Also, make sure you check your data types to ensure they make sense.

wf_psf/training/train.py
jeipollack: changed it.

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: This function was written by @tobias-liaudat. It's not used in training but for metrics analysis.  It looks like it will accept as input a tensor in tensor flow float units. but, I will know when I test it further when refactoring the `metrics` package.

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: I will make it a To Do.  As this module was written by Tobias, I was reluctant to review and refactor it as I didn't want to risk breaking sth.  We know it works as is.

wf_psf/training/train.py
jeipollack: fixed.

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: Replaced with Recursive Namespace.

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: added it to the To Do list to figure out what this comment is about.

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: added to the To Do list

wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: Thanks! Done.

jeipollack:
hi @sfarrens I believe I've addressed all your comments and those by @tobias-liaudat on change some config variable names. Re-ran the code and all is good.

sfarrens (approved):

PR #62, opened 2023-05-10
Refactor metrics

jeipollack:
This PR is more a refactoring of the training part of WaveDiff to remove redundant steps during metrics evaluation.  

Changes to be reviewed are in the following commits:

- [x]  https://github.com/CosmoStat/wf-psf/commit/40d5d0543f96fc1f787b3bfa3298157d1bd6e0b1

- [x]  https://github.com/CosmoStat/wf-psf/commit/d1a6e12510decea235b79d8803a77703d57d4778
- [x] https://github.com/CosmoStat/wf-psf/commit/1f273b114b82dac63a58a7346ff920e9f9a0694d
- [x] https://github.com/CosmoStat/wf-psf/commit/262801ed83922754f652f523f9a4db1513333450
- [x] https://github.com/CosmoStat/wf-psf/commit/58131f4bb648288cf0d41b789789bb081b8a7e85

I've performed other changes in the `tests` dir, `metrics_refactor.py`, etc. But, they're not ready for review, yet. So, this branch can't be merged.


jeipollack:
This last batch of commits can also be reviewed.

jeipollack:
one other comment, please don't delete branch bc I am still working on it

config/metrics_config.yaml
```diff
@@ -0,0 +1,118 @@
...
+    sed_interp_pts_per_bin: 0
+
+    # SED extrapolate 
+    sed_extrapolate: True
+    
+    # SED interpolate kind
+    sed_interp_kind: linear
+
+    # Standard deviation of the multiplicative SED Gaussian noise.
+    sed_sigma: 0
```
sfarrens: I know this is outside the scope of the review, but these parameters appear to be defined twice.

wf_psf/training/train.py
```diff
@@ -440,7 +400,9 @@ def train(training_params, output_dirs):
...
     final_time = time.time()
     logger.info("\nTotal elapsed time: %f" % (final_time - starting_time))
+    logger.info("\n Training complete..")
 
-    # Close log file
-    logger.info("\n Good bye..")
-    # log_file.close()
+    return (
+        psf_model,
+        training_handler._filepath_chkp_callback(checkpoint_dir, current_cycle),
```
sfarrens: If `_filepath_chkp_callback` is called outside of the scope of `training_handler` then it should probably not be a "hidden" attribute (i.e. no preceding `_`).

wf_psf/training/train.py
```diff
@@ -17,9 +17,9 @@
 import logging
 import wf_psf.utils.io as io
 from wf_psf.psf_models import psf_models, psf_model_semiparametric
-import training.train_utils as train_utils
+import wf_psf.training.train_utils as train_utils
 import wf_psf.data.training_preprocessing as training_preprocessing
-from wf_psf.data.training_preprocessing import TrainingDataHandler, TestDataHandler
+from wf_psf.metrics.metrics_refactor import evaluate_model
```
sfarrens: 1. Is `metrics_refactor` a temporary name while you are working on this module?
2. I don't see any calls to `evaluate_model` in this module, is it currently needed here?

config/data_config.yaml
```diff
@@ -0,0 +1,41 @@
...
+      SEDS: null
+      zernike_coef: null
+      C_poly: null
+      parameters:
+        d_max: 2
+        max_order: 45
+        x_lims: [0, 1000.0]
+        y_lims: [0, 1000.0]
+        grid_points: [4,4]
+        max_wfe_rms: 0.1
```
sfarrens: I think GitHub is complaining because there is no empty line at the end of the file.

sfarrens (changes requested):

sfarrens:
Hi @jeipollack here are my first comments on this PR. I will run some tests and get back to you if I have any further comments.

config/metrics_config.yaml
jeipollack: yes, it is a duplication due to copying from training_config.yaml. I am still working on the metrics configs. so, this isn't the final version and not meant to be part of this PR.

wf_psf/training/train.py
jeipollack: yes, it is temporary name.  And it was called but I decided to call it instead in `run.py`.  I will rm this import

jeipollack:
Everything is working as it should.  But, I see some possible configurations that could break the code or cause the user to not get the results they intended.

wf_psf/training/train.py
jeipollack: Fixed. I actually moved it outside of the class bc it's needed by the metrics module. I didn't see the need to make it a static method.

config/data_config.yaml
jeipollack: sorry I don't understand. how do you know it's complaining?

jeipollack:
The scope of this PR is to check the UI with the refactored metrics option.  I know the name `metrics_refactor.py` is not a good one, but all the commands in there were in `script_utils.py`.   The `metrics_refactor.py` is like an interface module for the `metrics.py` module, so it's not meant to override it.  Should I change the name to `metrics_interface.py`?

jeipollack:
To add, I realised that `run.py` has a bit more complicated because of the different config options.  I am thinking about how to make it cleaner/less complicated using factory methods.

config/data_config.yaml
sfarrens: I am pretty sure that's what the red circle with the minus means.
...............👇
image

wf_psf/run.py
```diff
@@ -78,15 +80,110 @@ def mainMethod():
...
+            data_params = read_conf(os.path.join(args.repodir, conf.data_conf))
+            logger.info(data_params)
+
         if hasattr(conf, "training_conf"):
             training_params = read_conf(os.path.join(args.repodir, conf.training_conf))
             logger.info(training_params.training)
 
+        if hasattr(conf, "metrics_conf"):
+            metrics_params = read_conf(os.path.join(args.repodir, conf.metrics_conf))
+            logger.info(metrics_params.metrics)
```
sfarrens: [STESS TEST] Providing an invalid path/filename or data type for any of the conf files provides an output log without any error information. The error can only be retrieved from the stderr.

wf_psf/run.py
```diff
@@ -78,15 +80,110 @@ def mainMethod():
 
     configs = read_stream(os.path.join(args.repodir, args.conffile))
```
sfarrens: [STRESS TEST] Providing an invalid key to `configs.yml` (e.g. `xetrics_conf` instead of `metrics_conf`) does not raise any error. The code simply runs as if `metrics_conf` was not provided. This doesn't break anything, but it may be worth adding a check that only valid keys have been provided.

sfarrens: [STRESS TEST] The config file is read in from a path relative to the repository directory. This only makes sense if you assume that the user will only ever use config files already provided in the repo. Otherwise, defining an absolute path to the config files can get complicated if `repodir` is anything other than `.`.

sfarrens:
> The scope of this PR is to check the UI with the refactored metrics option. I know the name `metrics_refactor.py` is not a good one, but all the commands in there were in `script_utils.py`. The `metrics_refactor.py` is like an interface module for the `metrics.py` module, so it's not meant to override it. Should I change the name to `metrics_interface.py`?

Yes, I think that would make more sense.

sfarrens:
@jeipollack I have added some feedback following some basic stress tests. You don't have to implement the changes in this PR, but I do think these points will be need to be addressed at some point.

config/data_config.yaml
jeipollack: I see. I though it was shifted by a row not column :). Fixed.

wf_psf/run.py
jeipollack: sorry does the data type have to do with the contents of the file or the file extension?

jeipollack: I ask bc if it's the file extension then the error would be FileNotFound

jeipollack: could you provide me an example of the test cases you used? (this is what the automatic testing program in Google wouldn't reveal :-) )

wf_psf/run.py
jeipollack: I corrected this with an exception

jeipollack: same is true about the data sets.

jeipollack: I'm thinking what to do.

sfarrens: Why not make the config file argument a full path in the first place?

wf_psf/run.py
sfarrens: Yes, sorry. What I meant is that if you did something like:

```yaml
data_conf: config/data_config.yaml
training_conf: config/training_config.yaml
metrics_conf: null
```

Where obviously you would expect an error for the invalid data type for the key `metrics_conf` this error only appears in the stderr and not in the run log.

wf_psf/run.py
jeipollack: I am thinking to change it in such way, but using the path to the config provided at runtime.

jeipollack: Ii changed it to full path for config file.  In the `data_config.yaml` file, I added an entry `data_dir` in both the `training:` and `test:` sections to specify the respective paths to train and test data in case they're stored in separate directories.

jeipollack: Just to add. the user needs only specify the name of the configs file in the `configs.yaml` file.

wf_psf/run.py
jeipollack: Hmm...not sure how to interpret this.  If the user entered `null` I would think it's because they only want to train the model. So, why should there be an error?

jeipollack: To test this quicker, I ran with the configuration below.  It throws a TypeError exception with the warning Training & Metrics run is skipped but carries on with metrics evaluation.

```
---
  data_conf: data_config.yaml
  training_conf: null
  metrics_conf: metrics_config.yaml
```

sfarrens: That's a valid point, but at it stands this will raise an error because it is looking for a path following the keywords. You could make `null` a valid data type, but what if someone puts e.g. `3`, the error should be logged so that the user know what went wrong with their config file settings.

jeipollack: if someone makes a typo in providing the config file name then an exception is thrown and output to log.  Would that be sufficient?

jeipollack: sorry I didn't refresh and saw your message above.  I will edit the code for this type of error.

jeipollack: okay I added Exception Handling for `TypeError` & `FileNotFoundError`.   Program should exist for those cases. Exception is `Null`.

jeipollack: (I kinda like writing this type of code) :D

jeipollack:
Latest modification (untested fully on JZ) is to store outputs in a unique dir `wf-outputs-YearMonthDayHourMin`.

```
wf-outputs
├── wf-outputs-202305261923
│   ├── checkpoint
│   ├── log-files
│   ├── metrics
│   ├── optim-hist
│   └── plots
├── wf-outputs-202305261925
│   ├── checkpoint
│   ├── log-files
│   │   └── wf-psf_202305261925.log
│   ├── metrics
│   ├── optim-hist
│   └── plots
└── wf-outputs-202305261926
    ├── checkpoint
    ├── log-files
    │   └── wf-psf_202305261926.log
    ├── metrics
    ├── optim-hist
    └── plots
```

I forgot to copy the config files over.  TO DO.

jeipollack:
Okay, I think I've implemented enough keys changes to close out this PR. 


I will save completing the unit tests for the next one.

```
└── wf-outputs-202305271930
    ├── checkpoint
    │   ├── checkpoint
    │   ├── chkp_callback_poly-coherent_euclid_200stars_cycle1.data-00000-of-00001
    │   ├── chkp_callback_poly-coherent_euclid_200stars_cycle1.index
    │   ├── chkp_poly-coherent_euclid_200stars_cycle1.data-00000-of-00001
    │   └── chkp_poly-coherent_euclid_200stars_cycle1.index
    ├── config
    │   ├── configs.yaml
    │   ├── data_config.yaml
    │   ├── metrics_config.yaml
    │   └── training_config.yaml
    ├── log-files
    │   └── wf-psf_202305271930.log
    ├── metrics
    │   └── metrics-poly-coherent_euclid_200stars.npy
    ├── optim-hist
    │   └── optim_hist_poly-coherent_euclid_200stars.npy
    └── plots
    ```

File: wf_psf/utils/io.py [hunk 7]
```diff
@@ -29,9 +30,19 @@ class FileIOHandler:
...
-        self._repodir_path = repodir_path
-        self._output_path = output_path
-        self._wf_outputs = "wf-outputs"
+        self.repodir_path = repodir_path
+        self.output_path = output_path
+        self._timestamp = self.get_timestamp()
+        self._parent_output_dir = "wf-outputs"
+        self._run_output_dir = (
+            self.output_path
+            + self._parent_output_dir
```
sfarrens: You are missing a `"/"` here. As it stands this raises an error as the directory cannot be created. 

Perhaps it would be safer/cleaner to use `os.path.join`.

wf_psf/run.py
sfarrens: Looks good now, good job!

wf_psf/run.py
sfarrens: This does not appear to be fixed, at least not as I expected. Misspelling the keys `training_conf` or `metrics_conf` is simply ignored.

sfarrens: OK, with the [latest commit](https://github.com/CosmoStat/wf-psf/pull/62/commits/70f1572d9616f619b74d6477e39d255391c33745) this is workable. I think we may want to revisit the error messages and exception handling in a follow up PR, but let's leave it for now.

sfarrens: OK, this seems to be mostly fixed. I think the only thing that is missing is an exception that is raised if the user tries to provide a path to `configs.yaml`, and it should be clearly specified that all the config files will have to be stored in the same location. Again, I think this can be addressed in a follow up PR.

sfarrens (approved):

sfarrens:
OK @jeipollack I am merging this PR. Good job! 👏

PR #63, opened 2023-05-31
Refactor plotting

jeipollack:
In this PR, I made the change of reducing some commands in `run.py` when reading in the `configs.yaml` file.  I tested that the `training` and `metrics` modules still run singly or jointly as well as error messages for incorrect entries in the config file.


jeipollack:
going to cancel this PR as I've done more refactoring to the UI and `run.py` in another branch `refactor_configs_run`

PR #70, opened 2023-06-26
Refactor plotting

jeipollack:
For this PR, the UI config handling for wavediff has been refactored using factory methods and the plotting has been enabled with `plots_interface.py`.

Now only a single parameter needs to be specified per run in the `configs.yaml` file. 

 ```
 ---
 training_conf: training_config.yaml
  ```
  
  The respective config files for `training` & `metrics` (and shortly plotting), now contain parameters for the other configuration files which ConfigHandler will use to determine the type of commands to execute during the run.
 
 Note, the code as is could launch multiple runs in serial with different configs files. To do this, just list all the config files in `configs.yaml` file with the correct key `training_conf`, `metrics_conf` or `plotting_conf` .

 I implemented various exception handling for `FileNotFoundError`, `TypeError` and `KeyError`.  I skipped `AttributeError` as I should write a function to check all the keys in the config file for typos/errors.


config/metrics_config.yaml
```diff
@@ -6,12 +6,14 @@ metrics:
   # Provide path to saved model checkpoints.
   chkp_save_path: checkpoint
   # Fill model_params if computing metrics_only on a pre-trained model
-  # ID name
+  # ID name  <-- this should be retrieved from the trained model training_config.yaml file
   id_name: -coherent_euclid_200stars
   # Path to Trained Model
-  trained_model_path: /Users/jenniferpollack/Projects/wf-outputs/wf-outputs-202305262344/
+  trained_model_path: /Users/jenniferpollack/Projects/jp272790/wf-outputs-202305272046/
```
sfarrens: Do you want push hard coded paths here? Not really a problem but you could just put `<PATH>` or something to let the user know this has to be specified.

config/plotting_config.yaml
```diff
@@ -0,0 +1,16 @@
+plotting_params:
+  # Specify path to parent folder containing wf-psf metrics outputs for all runs, ex: $WORK/wf-outputs/
+  metrics_output_path: /Users/jenniferpollack/Projects/jp272790/
```
sfarrens: Same point re: paths.

config/configs.yaml
```diff
@@ -1,4 +1,2 @@
 ---
-  data_conf: data_config.yaml
   training_conf: training_config.yaml
-  metrics_conf: metrics_config.yaml
```
sfarrens, line 4: I am a little confused by the current hierarchy of config files. Can you explain a bit about how this should work?

main_metrics.ipynb
sfarrens, line 1: Perhaps this could go in a e.g. "Notebooks" folder or something rather just at the root of the repo.

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+maintainers = [
+    { "name" = "Jennifer Pollack", "email" = "jennifer.pollack@cea.fr" },
+]
+
+description = 'A software framework to perform Differentiable wavefront-based PSF modelling.'
+dependencies = [
+    "numpy>=1.19.2",
+    "scipy>=1.5.2",
+    "tensorflow-addons>=0.12.1",
+    "tensorflow-estimator>=2.9.0",
```
sfarrens, line 18: Missing TensorFlow

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+  "pytest-pydocstyle",
+  "pytest-raises",
+  "pytest-xdist",
+]
+
+# Install for development
+dev = ["wf_psf[docs,lint,release,test]"]
+
+[project.scripts]
+myscript = "wf_psf.run:mainMethod"
```
sfarrens: Maybe go ahead and rename this to something useful.

src/wf_psf/psf_models/psf_model_semiparametric.py
```diff
@@ -27,6 +27,8 @@ class TF_SemiParam_field(tf.keras.Model):
 
     Parameters
     ----------
+    ids: tuple
+        A tuple containing a str id the PSF model class
```
sfarrens: You may want to double check the wording here.

src/wf_psf/utils/read_config.py
```diff
@@ -107,6 +105,8 @@ def read_conf(conf_file):
         except yaml.scanner.ScannerError as e:
             logger.exception("Improper syntax in yaml file.")
             exit()
+        except TypeError as e:
```
sfarrens: Why do you need the alias `e` here?

src/wf_psf/utils/read_config.py
```diff
@@ -123,16 +123,16 @@ def read_stream(conf_file):
 
     Yields
     ------
-    RecursiveNamespace
-        RecursiveNamespace object
+    dict
+        dict
```
sfarrens: Not very descriptive output

src/wf_psf/metrics/metrics_interface.py
```diff
@@ -24,6 +24,20 @@
...
+def ground_truth_psf_model(metrics_params, coeff_matrix):
+    psf_model = psf_models.get_psf_model(
+        metrics_params.ground_truth_model.model_params,
+        metrics_params.metrics_hparams.batch_size,
+    )
+    psf_model.tf_poly_Z_field.assign_coeff_matrix(coeff_matrix)
+
+    _ = psf_model.tf_np_poly_opd.alpha_mat.assign(
+        np.zeros_like(psf_model.tf_np_poly_opd.alpha_mat)  # type: ignore
+    )
```
sfarrens: What is the point of this `_` object?

src/wf_psf/metrics/metrics_interface.py
sfarrens, line 1: It seems to me that there is a good be of repetition in this module. It doesn't need to be resolved now, but it seems (at a quick glance) that more in-depth refactoring could be done.

src/wf_psf/plotting/plots_interface.py
```diff
@@ -0,0 +1,496 @@
...
+        "font.size": 16,
+    }
+    mpl.rcParams.update(plot_style)
+    # Use seaborn style
+    sns.set()
+
+
+def make_plot(
+    x,
+    y,
```
sfarrens: In general, I am against single character variable names as they make interpretation extremely difficult. I would put e.g. `x_axis` etc. Not critical, but worth thinking about.

src/wf_psf/plotting/plots_interface.py
```diff
@@ -0,0 +1,496 @@
...
+            A str representing dataset type, i.e. test_metrics or train_metrics.
+
+        Returns
+        -------
+        metrics_id: str
+            A str representing the id of a metrics for a given run.
+        rmse: list
+            A list containing root-mean-square (rms) errors for a metric for each run input.
+        std_rmse: list
+            A list containing standard errors on the root-mean-square (rms) quantites for a metric for each run input.
```
sfarrens: You can't have more than one return type. In this case, you are actually returning a tuple.

src/wf_psf/run.py
```diff
@@ -0,0 +1,108 @@
...
+
+def setProgramOptions():
+    """Define Program Options.
+
+    Set command-line options for
+    this program.
+
+    Returns
+    -------
+    args: type
```
sfarrens: Type `type`?

src/wf_psf/tests/metrics_test.py
```diff
@@ -0,0 +1,287 @@
...
+            training_params.id_name,
+            cycle,
+        )
+    )
+
+    poly_metric = metrics_handler.evaluate_metrics_polychromatic_lowres(
+        psf_model, simPSF_np, test_dataset
+    )
+    print(poly_metric)
+    print(paper_metrics["test_metrics"]["poly_metric"])
```
sfarrens, line 115: What's the point of these print statements in the tests?

sfarrens (commented):
Hey @jeipollack here is the first round of comments on this PR. I will take a deeper look at some of the modules once these points have been addressed.

src/wf_psf/tests/metrics_test.py
jeipollack: This was just some checks for myself.

Please do not review as I'm currently working on it.

config/metrics_config.yaml
jeipollack: I don't, but I have to for the time being.

Replaced.

config/plotting_config.yaml
jeipollack: Replaced.

config/configs.yaml
jeipollack: **Disclaimer:** Last Friday, just as I was explaining the latest implementation below I discovered that WaveDiff was not running in the way I expected, so I had to make it so! Along the way to do that I discovered I missed testing some additional configuration of train+metric+plot (compute/add metrics from other models), found a bug, and so that what's delayed me by a few days.  It's mostly sorted. In this new hierarch of config files I am missing one use case which I will explain at the end.

**Latest Changes**
The current implementation (which could be revised) is that WaveDiff can run a sequence of tasks for different purposes (ex: train two or more different sets of models) which are specified in `configs.yaml`.  

Ex: Running multiple training runs in sequence
```
---
  training_conf_1: training_config_1.yaml
  training_conf_2: training_config_2.yaml
  ...
  training_conf_n: training_config_n.yaml
```

[L83](https://github.com/CosmoStat/wf-psf/blob/8adb98ea26d26067056ce0e7bc6b3fb1a4863310/src/wf_psf/run.py#L83) will iterate through each of the config files and carry out the task specified by extracting the class id from the `key` in the config file, i.e. the training class id `training_conf` from `training_conf_1`.
If the user wants to train a model and run metrics evaluation on that model, then inside the `training_config.yaml` file at [L7](https://github.com/CosmoStat/wf-psf/blob/5e0b8fbc19fbac31d947f9074ff135b507eb04b4/config/training_config.yaml#L7) they should provide the name of the `metrics_config.yaml` file if in the same directory or the path `./path/metrics_config.yaml`  if in a sub-directory in the parent folder.
If the user wants to train, run metrics, and plot in a single run, then inside the `metrics_config.yaml` file at [L14](https://github.com/CosmoStat/wf-psf/blob/3042d92ac174439f7407f20b481706d2e7e8a3ef/config/metrics_config.yaml#L14) enter the name of the config file for plotting `plotting_config.yaml`.

Back to `configs.yaml` file, the user can mix training and metrics tasks as shown below but they are run independently of each other. 

Ex: Running multiple training runs + metrics run in sequence
```
---
  training_conf_1: training_config_1.yaml
  training_conf_2: training_config_2.yaml
  ...
  metrics_conf_1: metrics_config_1.yaml
  ...
  metrics_conf_n: metrics_config_n.yaml
```

This setup will be useful for when we implement parallelisation by parallelising the loop inside `run.py`.  

One missing use case is where the user wants to produce plots from with the outputs from these independent tasks. Something to think about later.

main_metrics.ipynb
jeipollack: Moved it.

pyproject.toml
jeipollack: Done.

config/configs.yaml
jeipollack: If we don't want to parallelise this way then instead of passing `configs.yaml` the code could be tweaked to receive any config file, i.e. `training_conf_xxx.yaml`, `metrics_conf_yyy.yaml` , or `plotting_conf_zzz.yaml`.   The key id to for the instantiating the right class would just be extracted from the input config file.

pyproject.toml
jeipollack: thx. Added.

src/wf_psf/psf_models/psf_model_semiparametric.py
jeipollack: Modified wording 

```
 ids: tuple
        A tuple storing the string attribute of the PSF model class
```

src/wf_psf/utils/read_config.py
jeipollack: Because, I did copy/paste without thinking. Removed.

src/wf_psf/utils/read_config.py
jeipollack: Updated the docstring desc.

src/wf_psf/metrics/metrics_interface.py
jeipollack: Yes, this could be further abstracted.

src/wf_psf/plotting/plots_interface.py
jeipollack: Oh kay.  Corrected.

src/wf_psf/plotting/plots_interface.py
jeipollack: Thanks for the feedback. I will keep it mind.

src/wf_psf/metrics/metrics_interface.py
jeipollack: This is Tobias' coding.  I see that `psf_model.tf_np_poly_opd.alpha_mat` is an instance of the class [tf.Variable ](https://www.tensorflow.org/guide/variable?_gl=1*4dfddg*_ga*MTMyMTIwNzQ1MS4xNjk0ODA4NTkz*_ga_W0YLR4190T*MTY5NTAyNDYyNy4yLjEuMTY5NTAyNDYyNy4wLjAuMA..).  It looks like Tobias was thinking that calling the class method `assign` required a dummy return object.  It's not needed.

src/wf_psf/run.py
jeipollack: updated the docstring

config/configs.yaml
sfarrens: OK, got it, thanks! This sounds good to me.

main_metrics.ipynb
sfarrens: Did you remove this file?

src/wf_psf/utils/read_config.py
sfarrens: Good, but it looks like you have the same situation on the exception above. Perhaps you can check them all just to see if you actual need/use the aliases.

sfarrens:
Hi @jeipollack sorry for the delay. There are just a couple of threads left to be resolved and then I believe we can merge this PR. 🙂

main_metrics.ipynb
jeipollack: Yes, I did.

src/wf_psf/utils/read_config.py
jeipollack: Hi, I made some changes to the exception handling in `run.py` and this `utils/read_config.py` module.  I rm the exception handling in the generator function `read_stream()` bc it doesn't do anything and won't log bc it's caught before log file is setup. For exception handling in generators, I read it requires a wrapper function that looks little complex (a nice exercise for the future). In the end, I just decided to catch the error in `read_stream` in `run.py`.  I tried some test cases that I've captured with different config files `config_typo.yaml`, `config_scanner_error.yaml`, `config_empty.yaml`, etc. In the next milestone I will integrate these tests more properly.

sfarrens (approved):

PR #71, opened 2023-06-29
Fix SimPSFToolkit import with sims module

tobias-liaudat:
There's an import problem causing the code to fail. The sims module is missing, this PR fixes it.

### DIFF
--- a/wf_psf/data/training_preprocessing.py
+++ b/wf_psf/data/training_preprocessing.py
@@ -9,7 +9,7 @@ import numpy as np
 import wf_psf.utils.utils as utils
 import tensorflow as tf
 import tensorflow_addons as tfa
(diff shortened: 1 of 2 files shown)


jeipollack:
Hi @tobias-liaudat, thanks for identifying this bug.  I would like to reproduce it can you let me know what settings one needs to do to replicate this failure?

tobias-liaudat:
@jeipollack if you do a clean installation in a new conda environment and then try to train the model. I used the branch `refactor_configs_run`. The command I used is the default one for the training:

`python wf_psf/run.py -c config/configs.yaml -r $WORK/repos/refactored_wavediff/wf-psf -o $WORK/outputs/refactor_wf`

PR #79, opened 2023-10-05
Feature end to end (Actually CI implementation)

jeipollack:
This PR includes a Github Actions workflow defined in `ci.yml`, which runs some unit tests for some methods in `io.py` module.  

There are also two validation tests for train.py and metrics.py, respectively. The baseline used for validation are the outputs of the main branch run over two cycles and epochs with a specific random seed used for initialisation of tensor flow.  These can't be run as unit tests because they require GPUs which are scarcely available when launching the CI.

To validate the training results, a mock run of the refactored code is launched using the same configuration as the main branch. Results are stored in the pytest fixture tmp_path_factory. The checkpoint weights and final psf models are read in and compared with the baseline results and rms is computed. The validation test asserts for each weight that the error < 1.e-9.

To validate the metrics, there are four tests to evaluate the metrics: low-res pixel reconstruction metric, optical path difference metric, monochromatic metric, and shape metrics produced by the refactored code with respect to the metrics computed by the main branch again ran over exactly the same configuration. The weights are read in and the psf model is regenerated and evaluated according to each metric. The tolerance level varies in this case where low res the tol is set to 1e-7 whereas the rest is 1e.-9. Bc I saw that the difference was ~4.8e-8 for the low-res pixel reconstruction.

I merged changes in the last PR to this branch. I will work on the end-to-end test in another branch.

Unused modules were removed for the release.

This PR resolves resolves issue https://github.com/CosmoStat/wf-psf/issues/47


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: CI-test
+
+on:
+  pull_request:
+    branches:
+      - dummy_main
```
sfarrens, line 9: Note that this will need to be changed when you are ready to merge `dummy_main` into `main`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: [ubuntu-latest]
+
+    steps:
+      - name:
+        uses: actions/checkout@v3
+      
+      - name: Set up Python 3.10.5
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10.5"
```
sfarrens, line 23: Why Python 3.10?

pyproject.toml
```diff
@@ -74,7 +74,6 @@ addopts = [
     "--verbose",
     "--black",
     "--emoji",
-    "--pydocstyle",
```
sfarrens, line 77: Why has this been removed?

requirements.txt
sfarrens, line 1: In principle, if all of these dependencies are covered in the `pyproject.toml`, then this file is no longer needed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+          
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
sfarrens: I am not sure I get the point here. If all the `wavediff` dependencies are listed in the `pyproject.toml` file then we should not need this line. If this `requirements.txt` file contains additional dependencies, what would be the situation where this file is not present (i.e. why the `if` statement)?

src/wf_psf/tests/data/validation/end_to_end/config/configs.yaml
sfarrens, line 1: It's best to avoid putting not Python files inside the package architecture if possible. These files will not be included in the built package by default.

sfarrens (commented):
Hi @jeipollack looks good overall. I think there are some files inside `src/wf-psf` that don't really belong there, but it should be fairly easy to fix.

.github/workflows/ci.yml
jeipollack: Because this is the version required by the version of Tensor Flow I'm using that doesn't break wavediff

pyproject.toml
jeipollack: I removed it bc it complained that there are no doc strings for the unit tests, which I read is not required for non-public modules by PEP-8.

.github/workflows/ci.yml
jeipollack: Thanks for pointing this out. I removes and replaced with the line to install dependencies for testing. ci test passed.

src/wf_psf/tests/data/validation/end_to_end/config/configs.yaml
jeipollack: Removed

requirements.txt
jeipollack: deleted

.github/workflows/ci.yml
jeipollack: yes, I'm aware.

sfarrens (approved):

PR #85, opened 2023-11-14
Feature unit tests run configs - MetricsConfigHandler Class

jeipollack:
Wrote a few unit tests for the `configs_handler.py` module and one for the `psf_models.py` module.
This lead to some code refactoring of the `MetricsConfigHandler` class, which introduced some bugs during the process in some commits (corrected).  

Unit tests passed and validation tests which check reproducibility under a couple of different configurations passed.

I would like this PR to go to `dummy_main`. I spotted missed unused code which I forgot to delete.


src/wf_psf/tests/test_utils/pytest.xml
sfarrens, line 1: If this is Pytest output you probably don't want to include this in your PR.

jeipollack: You’re right i don’t. i thought i rm them all. By mistake i clicked a generic commit button in VS code and it seemed to stage every new file or file that changed.

sfarrens:
@jeipollack once you resolve the thread above this PR can be merged

src/wf_psf/tests/test_utils/pytest.xml
jeipollack: I removed it.  I don't see any other `pytest.xml` files. is this thread you were speaking of?

sfarrens: Yep, that's it. You could add `pytest.xml` to your `.gitignore` so you don't track these files.

sfarrens (approved):

jeipollack:
I added pytest.xml yesterday to .gitignore but forgot to comment so

PR #94, opened 2023-12-08
Feature 92 ci ignore gpu tests

### LINKED ISSUE #92: Update CI workflow to ignore tests which require GPU
Validation tests for training and metrics evaluation are launched with `pytest`. However, because they require the usage of GPU they're not suitable to run during CI because GPUs available on Github-hosted runners are scarcely-available. 

In release-2.0.0., the validation tests are skipped with the `pytest.mark.skip` decorator.  Instead we can use the decorator `pytest.mark.skipif`, adding a conditional to be evaluated that will skip the test only during CI.

jeipollack:
This PR applies a fix raised in #92 regarding skipped validation tests.  We want the validation tests to be skipped during CI, but to be ran manually with `pytest` by the developer.  We applied the conditional `pytest.mark.skipif` decorator and will check if the CI environment variable was defined. If yes, the test is skipped, else the test is run.

Closes #92

### DIFF
--- a/src/wf_psf/tests/metrics_test.py
+++ b/src/wf_psf/tests/metrics_test.py
@@ -107,7 +107,7 @@ def main_metrics(training_params):
     return np.load(os.path.join(main_dir, metrics_filename), allow_pickle=True)[()]
 
 
-@pytest.mark.skip(reason="Requires gpu")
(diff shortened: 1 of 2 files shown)


jeipollack:
A comment:  I began working on this in the very late hours last Friday.  I made some stupid errors bc I was tired. So, I decided to do a hard reset to get rid of the many useless commits.

PR #97, opened 2023-12-11
Fix Copy-Paste Error in Epochs Settings

nadamoukaddem:
fix: correct duplicate 'n_epochs_non_params' in train.py

Please note that I couldn't test the changes as I currently don't have access to GPUs.


jeipollack:
Thanks, @nadamoukaddem. I did a test run with a different epoch settings for parametric and non-parametric training and confirmed the bug is now fixed.

Issue #114, opened 2024-02-05
Add Code_of_Conduct, Contributions and WaveDiff_Dev_Workflow markdowns


jeipollack:
I prepared three Google Docs which should be converted to Markdown files:

* [CODE_OF_CONDUCT.md](https://docs.google.com/document/d/130iB4VN3TY2wZXcN7MxaqikgPDmaJbDjJfSDPRAj_Tw/edit)
* [CONTRIBUTING.md](https://docs.google.com/document/d/1bUPLYbMjlcUseE2gZOaC5pEGBNTvR2-4UV3XVbfgIwI/edit)
* [DEV_WORKFLOW.md](https://docs.google.com/document/d/1SJyAw7tEBRcJTa62SkU3qvjgoxrtoPbqRSEjaTIPQjA/edit)

Save the files in the top-level of the repo.

Update the documentation file: https://github.com/CosmoStat/wf-psf/blob/main/docs/source/index.rst?plain=1 by adding at the bottom:

```
Contributing
------------

For information on how to contribute to the development of WaveDiff, see
`Contributing to WaveDiff <./CONTRIBUTING.md>`_.
```

Make sure the links work.

jeipollack:
Hi @nadamoukaddem I granted you edit access, but if there are changes you would like to propose please you suggest them instead of editing the document directly. Thanks

nadamoukaddem:
Hi @jeipollack , should I include the last paragraph in [DEV_WORKFLOW.md](https://docs.google.com/document/d/1SJyAw7tEBRcJTa62SkU3qvjgoxrtoPbqRSEjaTIPQjA/edit) ?

jeipollack:
> Hi @jeipollack , should I include the last paragraph in [DEV_WORKFLOW.md](https://docs.google.com/document/d/1SJyAw7tEBRcJTa62SkU3qvjgoxrtoPbqRSEjaTIPQjA/edit) ?

Hi @nadamoukaddem, thanks for catching this. Omit it from the new markdown file.

nadamoukaddem:
Hi @jeipollack , I added the files and updated the documentation file, but when I opened the pull request, I didn't pass the test with pytest. I'm also having the same issue when running the tests on #112.

PR #117, opened 2024-02-13
Add doc files 

nadamoukaddem:
solves #114. Additionally, I downgraded the pytest version in the pyproject.toml file so it can pass the tests with pytest.


pyproject.toml
jeipollack, line 1: Can you undo this change as it will be addressed in the other Pull Request? 

Even though the CI tests fail the scope of this PR are not related.

branching_model.png
jeipollack, line 1: Can you put this image in the assets directory? 

You will then need to correct the link in the DEV_WORKFLOW.md.

jeipollack (commented):
Even though the CI tests fail, the scope of this PR is just documentation.  
Let me know when you do the small changes so I can do the merge. Thanks.

branching_model.png
nadamoukaddem: I moved the image and corrected the link

pyproject.toml
nadamoukaddem: Done

nadamoukaddem:
I made the changes.

pyproject.toml
jeipollack: I don't see the commit where this was done.

nadamoukaddem: I should have added the change with this commit: `Moved image to assets directory`

jeipollack: I see it was part of the https://github.com/CosmoStat/wf-psf/pull/117/commits/104c4c1730a37d23e6014a363363b3ad54edbcb2

I've done such in the past, but best practice is that the commit should be specific to a scope.

nadamoukaddem: Thank you for pointing that out. I'll make sure to keep my commits more focused in the future.

Issue #133, opened 2024-04-03
Validation of Physical Layer Model with simulations 


jeipollack:
To validate the Physical Layer PSF model, we will use simulations that include a Zernike prior.  Metrics and plots should be generated and compared with those reported in Tobias' thesis (section 6).

The simulations are located on Jean-Zay: `/gpfswork/rech/ynx/commun/data/euclid_sims`.

- Check out a new branch from `case_study_psf_decontamination` to use for validation
- Update the configuration files: `data_config.yaml`, `training_config.yaml`, etc. with the configuration settings described in Tobias' thesis
- Run WaveDiff for each dataset
- Produce similar plots as found in the Chapter 6: Results section
- Evaluate whether the results resemble those in the thesis

nadamoukaddem:
What does 'SFE' stand for in the name of datasets, and are the 'nm' numbers for SEDs resolution?

jeipollack:
The answers (SFE=Surface Errors and XX nm mean WFE RMS are error maps that were used to build the Zernike prior) to your questions are explained in section 6.2 of Tobias' thesis.

nadamoukaddem:
Thank you, Jennifer. I am reading Chapter 6 from the beginning so I don't miss anything. I have some questions. 

- Tobias mentioned in his thesis that he used only the dataset with SFE. Why do we need the one with no_SFE?
-  What is the number of stars in these simulated datasets?
-  Are the plots to be produced only for the waveDiff-polygraph model? 
- Why do we have a different number of Zernikes in the training configuration (15) compared to the metrics configuration (45)? 
- The same goes for the number of bins: 8 for training and 20 in metrics. 
- What does the parameter M=64 mentioned in the thesis refer to?

jeipollack:
Hi Nada, it’s good that you read chapter 6 and aiming to develop some
understanding surrounding this task.

As I believe one of your goals is to do research, i want to remind you that
a fundamental part of being a researcher is “Reasoning”.

So, with your questions first try to determine the answer yourself. To aide
you, you can look at other parts of Tobias’ thesis by doing keyword
searches. For some questions the answer may be obvious, whereas for others
your reasoning may come with some uncertainty. It is those questions where
you think your reasoning is uncertain that you should ask for clarification.

The point is don’t just ask what is something or why. Try to determine this
yourself and present what you think along with your question.


I will answer one of your questions directly.

You need to run WaveDiff on the model with the physical layer. You will see
[...]

PR #120, opened 2024-02-19
Fix black test failure in CI 

nadamoukaddem:
this PR should solve #119. 

To address the issue, I temporarily set the pytest version to 7.4.4 and reformatted the code. Additionally, I upgraded the Python version in CI to 3.10.12.

PR #142, opened 2025-02-05
WaveDiff: New PSF Model Features, Refactoring, Rotation Obscuration Mask, and PEP8 Compliance

### LINKED ISSUE #89: Refactor train_utils module
A review of the [train_utils](https://github.com/CosmoStat/wf-psf/blob/main/src/wf_psf/training/train_utils.py) module revealed areas needing improving in terms of code quality.  There are duplicated commands, hard-coded values and many nested-if statements.

Refactoring should apply the following:

- Moving duplicated commands and if-statements either to individual functions, or a dictionary of options with specific function calls, or strategy design pattern to handle the different settings
- Convert hard-coded values to options or remove entirely

### LINKED ISSUE #90: Update to recent versions of TensorFlow
WaveDiff implements the Rectified Adam Optimiser from the TensorFlow Addons library that has stopped development (see details in [link](https://www.tensorflow.org/addons)). Minimal maintenance releases will continue until May 2024.  As a result, it is not compatible with the latest versions of TensorFlow 2.11+. Interestingly, 2.9+ is also affected resulting in the following error reported in Issue #88 which results when loading a saved checkpoint.  The Rectified Adam Optimiser is currently not part of the core library of TensorFlow 2.  The fix is use the `tf.keras.optimizers.legacy` namespace (see [here](https://www.gitclear.com/open_repos/tensorflow/tensorflow/release/v2.11.0)) that allows the old optimizers to work. 

This issue to do a couple of tasks:

- Re-evaluate performance with the Adam Optimiser versus the Rectified Adam Optimiser to see if it is necessary to use it
- If Adam Optimiser is good enough, update WF to recent versions of TF with new optimiser functionality 
- If not, re-evaluate possible next steps such as asking TF community if and when migration of Rectified Adam will happen and/or development of custom optimiser

### LINKED ISSUE #91: Update Phase Retrieval code in old WaveDiff with Random Seed setting
Validation of WaveDiff 2.0.0 has been done using the older version of the code in the branch [main_random_seed](https://github.com/CosmoStat/wf-psf/tree/main_random_seed ) that was upgraded to include the user setting of a fixed random seed.

Tobias' Phase Retrieval implementation in the old code should be upgraded with the fixed random seed in two commit states It's possible to cherry-pick one of the commit states in [tf_psf_field](https://github.com/CosmoStat/wf-psf/commit/10b88940c1d377ec7216fa42685c92c0e9c3dbcd) , but not the other in [long-runs/train_eval_plot_script_click_multi_cycle.py](https://github.com/CosmoStat/wf-psf/commit/a0510fd2f15cea2ab64c985917418311e135799c) as they include some other unnecessary changes (bad practice on my part).  But since they're really simple, so these changes could be added by hand.

Once Tobias updates his Phase Retrieval code (in the old version) it should be used to produce new baseline results to be used for validation.

### LINKED ISSUE #113: Optimise WaveDiff to address Centroid problem
During PSF model optimisation, the centroids between two images: reconstructed and actual must be the same. Else, the residual errors will be large even though the model is a good representation.

To enable better centroiding requires reintroducing a PSF model called [TF_physical_poly_field](https://github.com/CosmoStat/wf-psf/blob/10b88940c1d377ec7216fa42685c92c0e9c3dbcd/wf_psf/tf_psf_field.py#L512) which is a polychromatic PSF model with a physical layer.

### LINKED ISSUE #116: Reintroduce SpatialVaryingPSF module as a refactored class into v2.0+
For generating spatially varying PSFs to serve as validation data, we need to reintroduce the module from the old main branch: https://github.com/CosmoStat/wf-psf/blob/old_main_backup/wf_psf/GenPolyFieldPSF.py

The module should be refactored to improve:

- readability, i.e. update class names and methods
- isolate tasks to enable unit testing
- add doc strings for API documentation

### LINKED ISSUE #119: Failure of Black Test in CI testing in github
**Description of the bug**:
I opened a pull request to add some documentation files. However, during CI testing on GitHub, some .py files failed the black test (see image below and this [link](https://github.com/CosmoStat/wf-psf/actions/runs/7929828212/job/21650841152?pr=117)
 for the full log description), despite passing when I ran these tests on my Linux machine. Note that I didn't make any changes to these Python files in the PR.

pytest

**Expected behavior**:
 All files should pass the CI testing.

### LINKED ISSUE #123: Implement PSF model with a physical layer
To address various issues in the modelling like centroid problem, chip misalignments, and including a prior a new PSF model class needs to be added to WaveDiff >v2.0.x.  
 
It's implementation is illustrated in the far left in this image:
IMG_20240213_181157

The PSF model class is in version 1.0 but requires refactoring to be included in v2.0.x

### LINKED ISSUE #125: Bug: CI fails with pytest 8.1.dev
Related issue: https://github.com/pytest-dev/pytest/issues/11779 opened by `astropy` developer.

Response: https://github.com/pytest-dev/pytest/issues/11779#issuecomment-1879387034

> py.path -> pathlib.Path transition, the py.path hook parameters have been deprecated in the 7.x series, changed to an error in 8.0, and finally removed in 8.1. It seems like the pytest_ignore_collect hook implementation in the pytest_filter_subpackage plugin still uses the deprecated path argument, should use collection_path instead (available since pytest 7.0.0). 
>
> If you can require pytest>=7, then can just replace path -> collection_path (and possibly adjust for py.path.local -> Path).
> 
> If you still need to support pytest<7, you'll have to do something like the following I'm afraid:
> 
> if PYTEST_GTE_7:
>    def pytest_ignore_collect(collection_path):
>        ...
> else:
>    def pytest_ignore_collect(path):
>        ...
> 

```
Run python -m pytest
...
Argument(s) {'path'} are declared in the hookimpl but can not be found in the hookspec
[49](https://github.com/jeipollack/wf-psf/actions/runs/8139196676/job/22241824219#step:5:50)
Error: Process completed with exit code 1.
```

I saw someone add it to their `conftest.py` but I don't fully understand the implementation as it seems to go a bit deep into pytest packages and wrappers I'm unfamiliar with. The other solution I've seen is to lock `pytest <8.0.0` until needed pytest-plugin updates.  However, we know we can use `pytest==8.0.x`, so perhaps we do the same.

### LINKED ISSUE #132: Change random seed for the data-driven part every time it gets reset and re-initialised
When we are working in the phase retrieval configuraiton it is important that each time we re-initialise the data-driven part we use a different seed so that we can restart the optimisation from a different point in the optimisation landscape.

The reset is handled through the `TF_NP_poly_OPD.init_vars()` method that re-initialises the  data-driven part. However, the random seed is set to the config file random seed each time it is done.

If we're running several cycles we need to change the seed that is being set each time we re-initialise the data-driven part.

### LINKED ISSUE #139: Rename variable as psf_model instead of specific type
https://github.com/CosmoStat/wf-psf/blob/bfd1ef806e332dd9297dbaa2977d95cbc4ffbe72/src/wf_psf/training/train_utils.py#L46

jeipollack:
This PR includes new features:

* Including PSF model extension with physical layer for a Zernike prior (Closes #123)
* Centroid Estimation correction (Closes #113)
* CCD z-axis misalignment model correction 
* Add module for simulating spatially-varying PSFs and validation tests  (Closes #116 and #133)
*  Bug fix for random seed in data-driven PSF model during re-initialisation (Closes #132)
*  Add phase retrieval and validation tests (Closes #91)
* Replacing TrainingDataHandler and TestDataHandler with a single DataHandler class  
* Renaming variables, functions, and classes following PEP8 naming conventions (Closes #139)
* Removal of used imports and deprecated code 
* Replace black with ruff in `pyproject.toml` (Closes #119,  #120,#145)
* Unpin pytest in `pyproject.toml` (Closes #125)
*  Increase version of TensorFlow (2.11.0 or 2.12.0) not most recent (Closes #90)
* Refactors module `train_utils.py`  with helper functions and removes unused code (Closes #89 and #152)
* Add `test_training` directory with unit tests for `train.py` and `train_utils.py` modules
* Add Obscuration Mask Rotation option
* Refactoring of `train_utils.py` and `train.py` modules using Strategy Pattern
* Data Generation script for generating datasets with varying levels of complexity to use to validate this release and future releases of WaveDiff
* Add custom loss and metrics function to handle masks update new & existing functions for centroid estimation and sample weights when a mask is included in the dataset


jeipollack:
@sfarrens this PR is still in draft mode, but since it is rather long perhaps I can indicate which parts are ready for review. What do you think?

tobias-liaudat:
The last commits I pushed addressed Issue #162 by proposing a generalised sigmoid function to apply to the computed sample weights, allowing us to handle scenarios with more realistic noise distributions. The formula came from tests with real Euclid data. The final values to use should be determined by further testing on real data, although these should not be far from the current default ones. (Closes #162)

tobias-liaudat:
The last commits include a general script to simulate datasets that contain all the currently supported scenarios and should integrate all the previously created datasets (Closes #160 )

PR #150, opened 2025-03-05
Include mask handling for training and testing.

### LINKED ISSUE #143: Adapt WaveDiff training to account for masks
Real data will contain masks for cosmic rays so we want to avoid using this pixels for computing the loss and back-propagating the gradients.

I suggest starting with a (wavediff) simulated dataset and incorporate either simulated or real masks.

Validation TBD. One idea is verifying that training without masks or training with masks full of ones give the same results. 

Team: @tobias-liaudat @jeipollack @CentofantiEze

### LINKED ISSUE #146: Forced calculation of shape metrics on test dataset
Why are we forcing the calculation of shape metrics on test dataset? Shouldn't we also use the `metrics_params.eval_train_shape_results_dict` value from the config file?

The hardcoded `True` value can be seen [here](https://github.com/CosmoStat/wf-psf/blob/383726edc6c20d5136fb9e6f95c8af9fd57b5396/src/wf_psf/metrics/metrics_interface.py#L310).

CentofantiEze:
### Description:
Include a custom loss functions that allows training with masked data and accounts in the loss for the masked pixels which are not taken into account. 

### Content:

- Added function `masked_mse` which allow to compute the masked mean squared error.
- Added the `MaskedMeanSquaredError` loss class using the `masked_mse` function.
- Added the `MaskedMeanSquaredErrorMetric` metric class using the `masked_mse` function.
- Updated the `general_train_cycle` to handle the masked loss and metrics.
- Modified the `train()` function defined in `src/wf_psf/training/train.py` to handle the masked data as well as the masks.
-  Handling of the mask in the noise estimation algorithm.
- Configuration parameter for computing shape metrics for test dataset as optional

### Comments

The implementation has been validated by training wavediff with simulated masked stars. The results are commented in the issue #143.

Closes #143 
Closes #146


CentofantiEze:
Mask handling for noise estimation was added.

src/wf_psf/metrics/metrics.py
```diff
@@ -61,6 +62,9 @@ def compute_poly_metric(
         is present, the noiseless stars from the dataset are used to compute the metrics.
         Otherwise, the stars are generated from the gt model.
         Default is `None`.
+    mask: bool
+        If `True`, the predctions are masked to match the target masks.
```
jeipollack: I think the description could be expressed more clearly:

```
If `True`, predictions are masked using the same mask as the target data, ensuring that metric calculations consider only unmasked regions.
```

jeipollack:
Hi @CentofantiEze, amidst distraction I am starting to look at your PR. Could you also provide your test dataset in the `data/` file as well as your validation tests in `src/wf_psf/tests/data/validation/` as a new directory?

src/wf_psf/metrics/metrics.py
CentofantiEze: I agree, thank you. The mask docstring was updated.

CentofantiEze:
Hi @jeipollack I added the masked test dataset to the `data/masked_dataset/` folder. Should I also include the masked training dataset as well as the unitary masked train and test datasets?

jeipollack:
> Hi @jeipollack I added the masked test dataset to the `data/masked_dataset/` folder. Should I also include the masked training dataset as well as the unitary masked train and test datasets?

Hi @CentofantiEze, yes, please. I am going to refactor some parts of the code in `train.py`, and write a few unit tests before approving the PR.  I would like to be able to run your validation tests and reproduce your results as a sanity test.

jeipollack:
Thanks, @CentofantiEze. You can add a new directory here: https://github.com/CosmoStat/wf-psf/tree/feature-mask-loss/src/wf_psf/tests/data/validation with your config files in a new directory called `mask_handling` (maybe)...I hope it's not too much trouble.

CentofantiEze:
@jeipollack would you need the wf-outputs directory with the optim-hist, checkpoint, metics, etc., or just the config files? 

If you need the wf-outputs it might take me some time (jz is on maintenance) because I need to re-run the validation, since I've updated the training and testing datasets. If not, I have the config files for the three validation cases. 

- Training and testing on masked data.
- Training on masked data and testing on unitary masked data (effectively not masked).
- Training and testing on unitary masked data (control case, should be equal to not using the masks).

jeipollack:
> @jeipollack would you need the wf-outputs directory with the optim-hist, checkpoint, metics, etc., or just the config files?
> 
> If you need the wf-outputs it might take me some time (jz is on maintenance) because I need to re-run the validation, since I've updated the training and testing datasets. If not, I have the config files for the three validation cases.
> 
> * Training and testing on masked data.
> * Training on masked data and testing on unitary masked data (effectively not masked).
> * Training and testing on unitary masked data (control case, should be equal to not using the masks).

Hi @CentofantiEze, thanks for asking. I was thinking more of the configuration files.  But, now I realise it would be nice to have everything (especially the metrics) and your plot(s).

CentofantiEze:
Ok, I'll upload it as soon as JZ is back online and all the jobs run successfully.

CentofantiEze:
Added validation tests in `src/wf_psf/tests/data/validation/` containing the configuration files, jobs submission files, resulting metrics and opt-hist for each validation run. Also the results are shown in figure `masked_loss_validation.pdf`.

Screenshot 2025-03-14 at 10 20 35

config/training_config.yaml
```diff
@@ -126,7 +126,10 @@ training:
   # Training hyperparameters
   training_hparams:
     batch_size: 32
-         
+
+    # Loss function to use while training. Options are: 'mask_mse' and 'mse'. Default is 'mse'.
+    loss: 'mask_mse'
```
tobias-liaudat, line 134: The comment says that the default is "mse", but the example config says "masked_mse". It'd be good to change it to "mse"

src/wf_psf/training/train_utils.py
```diff
@@ -77,6 +77,74 @@ def on_epoch_begin(self, epoch, logs=None):
...
+    """Masked Mean Squared Error.
+
+    Parameters
+    ----------
+    y_true: Tensor
+        True values
+    y_pred: Tensor
+        Predicted values
+    mask: Tensor
+        Mask to be applied
```
tobias-liaudat: This is the "key" function of the masked part. We need to be more precise on the conventions we are using for the mask. It could lead to some confusion on future developments. For example, be precise if the mask is booleans or floats, and in the case of floats if it can take vales only on the set {0,1} or in the interval [0,1].

As I see it, the mask could be any float values as `unmasked_pixels = tf.reduce_sum(mask, axis=[1, 2])` is taking the  `mask_weight` or `mask_sum`. The name  `unmasked_pixels` might be confusing as it seems that the mask can only be 0s or 1s (thus counting pixels), but it doesn't need to be.

To avoid confusion I suggest:
- improve the docstring of `mask` to specify how the mask should be (e.g. floats not necessary 0s or 1s, where `0` would make the pixel not being taken into account)
- change the variable name `unmasked_pixels` by `mask_sum` or `mask_weigth` to avoid confusion

src/wf_psf/training/train_utils.py
```diff
@@ -195,14 +265,28 @@ def calculate_sample_weights(outputs: np.ndarray, use_sample_weights: bool) -> n
...
         
-        # Estimate noise standard deviation
-        imgs_std = np.array([std_est.estimate_noise(_im) for _im in outputs])
+        if loss == "masked_mean_squared_error":
+             images = outputs[..., 0]
+             masks = np.array(outputs[..., 1], dtype=bool)
+             masked_windows = std_est.window * masks
+             imgs_std = np.array(
+                 [
+                     std_est.estimate_noise(_im, _win)
```
tobias-liaudat: I see here that the signature for the `estimate_noise()` has changed with respect to the standard version of the method in the `NoiseEstimator` class, but the standard class has not changed, which would raise an error. It seems that the modifications to this class have not been pushed to the PR or are missing.

tobias-liaudat: The standard version only takes `image` as input and not a window function

tobias-liaudat (changes requested):
Everything looks good; I just left three minor comments to address.

Thanks for this PR!

config/training_config.yaml
jeipollack: Thanks, for noticing.  Do you think the default should be `mse`? I think it is more likely with real observations to expect a mask, no?

tobias-liaudat: We can focus on real data and leave it on "mask_mse", but we should then change the docstring specifying the default, right?

src/wf_psf/training/train_utils.py
jeipollack: Hi @tobias-liaudat, I updated the doc string with type hints.  If we define these terms as `tf.Tensor` we can avoid enforcing floats vs. ints.  I clarified in the docstring the range of acceptable values for the mask [0,1).  I also replaced `unmasked_pixels` with `mask_sum`.

Let me know if these changes are what you had in mind.

tobias-liaudat: Looks better, thanks!

src/wf_psf/training/train_utils.py
jeipollack: As discussed in our huddle, this was a rebase error. I corrected it and also added a few unit tests.

Bug was missed during training because `use_sample_weights=False`.  I will update the existing unit tests to cover the different settings.

tobias-liaudat:
Ah, sorry to bother you with one last comment, @CentofantiEze. I think it'd be good to add the script to generate the masked dataset (and eventually the job associated with it) in this PR so that there's a track of how that dataset was generated.

It's important to have a track of how things where created for future tests.

CentofantiEze:
Hi @tobias-liaudat yes I forgot to add the notebooks I used for generating the masked datasets. I uploaded the notebooks to the directory which contains the masked datasets.

tobias-liaudat:
Thanks @CentofantiEze !

config/training_config.yaml
jeipollack: Modified. If you agree with changes could you mark this comment as resolved?

src/wf_psf/training/train_utils.py
jeipollack: Thank you as well. There was a typo in the return statement, I corrected it.

src/wf_psf/training/train_utils.py
jeipollack: Unit tests have been updated but I decided to propagate the masked_windows directly to the NoiseEstimator class with a new method called `apply_mask`. It seemed more natural that this check should happen there rather than in `calculate_sample_weights`. Could you review these changes? I am doing a full test run with these changes including the `use_sample_weights=True` and `False`.

jeipollack:
@tobias-liaudat I also updated the `metrics_config.yaml` file with the option to trigger or not the shape metrics pipeline  for the test dataset removing the hard-coding.  This is propagated to `metrics_interface.py' file.

tobias-liaudat:
@jeipollack Should I go over the code again?

jeipollack:
> @jeipollack Should I go over the code again?

Yes, @tobias-liaudat  could you please look over the changes?

jeipollack:
My runs are still going. I think if the results are reproducible and the `use_sample weight=True` results make sense and you're satisfied with the changes, then we can merge.

jeipollack:
@tobias-liaudat Ready for you to review. 

Here's a plot of the results of Ezequiel's original results, where the case for ('mask_train_mask_test') was reproduced, and I added the result when the `use_sample weight=True`  and masking applied in the Noise Estimation. Results are consistent within the error.

masked_loss_validation_with_sample_weights-2.pdf


If you're satisfied, we can proceed with merging.

tobias-liaudat (approved):
It looks great to me. Thanks for the work, @CentofantiEze @jeipollack !

I'm happy to approve this PR

jeipollack:
Thanks @tobias-liaudat! Can you do the merge, since you became the final reviewer?

PR #166, opened 2025-06-21
165 refactor tfphysicalpolychromatic to cleanly separate training vs inference behaviour

### LINKED ISSUE #165: Refactor `TFPhysicalPolychromatic` to cleanly separate training vs inference behaviour
When loading a trained model for inference, the `TFPhysicalPolychromatic` class still requires access to the dataset used for training, including observed positions. This causes unnecessary coupling and can break inference pipelines.

Code in question:
https://github.com/CosmoStat/wf-psf/blob/f15c0f2068a48f24ae1962cd1791de31989f7c28/src/wf_psf/psf_models/models/psf_model_physical_polychromatic.py#L77C7-L77C35

## Problem

Currently, instantiating a trained PSF model for inference fails unless the original data object (including observed positions) is provided. This happens even when config flags such as correct_centroids: False are set.

**Specific issues:**

- `self.obs_pos` is computed and passed into a TFPhysicalLayer even during inference.
- `self.zks_prior` is computed via `get_zernike_prior(...)`, which always queries the full training/test dataset—including centroid correction and CCD misalignment contributions—even if those options are disabled.
- Even with `correct_centroids: False`, the call to `get_obs_positions()` is still executed.

## Reason
Originally, the class was tightly coupled to training-time behavior, as described in this Slack comment by Tobias:

"The original idea of the Zernike prior was that you need the prior at the positions you’re going to compute the PSF. That motivated the development of the physical layer. Then we extended that layer to handle the centroids, CCD misalignments, etc. [...] The issue is that the position is used for indexing as it allows to gather the correct priors in the Zernike contribution list."

This design works for training, but breaks modularity and hinders clean inference from pre-trained weights.

## Desired behaviour
Inference mode should:
- Not require loading the full training dataset.
- Be decoupled from training-time artifacts like centroid correction and full dataset access.
- Support inference at arbitrary positions, using:
  - A precomputed Zernike prior (if used), matched to inference positions.
  - Optional CCD misalignment correction (computed on the fly using inference positions).
  - SED and position inputs at prediction time.

Training mode should:
- Retain full control over dataset access and augmentation.
- Aggregate Zernike contributions (including centroids and CCD misalignment).

## Proposed solution
Refactor the class to distinguish clearly between **training** and **inference** use cases.

- Support lazy loading for positions and physical layer loading, and possibly even the prior but during model creation.
- Avoid calling dataset-related methods during inference unless explicitly requested.
- Introduce a factory method for inference setup

jeipollack:
Closes #165 

This PR refactors the Physical Layer Polychromatic PSF model to separate training from inference settings, enabling more flexible and modular usage across different execution contexts.

### Core Refactoring
- **New `run_type` attribute**: Added to `data_handler` supporting `training`, `simulation`, `metrics`, and `inference` modes
- **`ZernikeInputsFactory` class**: New factory class that builds `ZernikeInputs` instances based on the specified run type
- **Enhanced `ZernikeInputs` data class**: Now intelligently assembles Zernike contribution lists based on `run_type` and available data (including priors and source positions for CCD misalignment corrections)

### Additional Improvements
This PR expanded in scope to include several other enhancements:
- **Hybrid loading pattern**: Refactored to use eager loading in constructors combined with lazy-loading via property decorators, reducing constructor overhead and improving modularity
- **Code organization**: Improved module structure including new `tf_utils.py` module in the `psf_models` sub-package 
- **Bug fix**: Fixed incorrect argument name in `DataHandler` class that prevented training dataset images from being converted to TensorFlow data type as expected
- **Helper Methods**: Added several helper methods to assist with preparing datasets and ensure data type conversion is performed as expected
- **Deprecations**: Removed `get_obs_positions` as it was no longer needed

## Validation

Comprehensive validation was performed across multiple datasets (v0xx and v1xx) with comparisons against `case_study` and `main` branches.

**Results**: [[Google Sheets](https://docs.google.com/spreadsheets/d/1jIyvXOCuHDY8P548oaWMtbU2hrLgxuKEvgL8D92dNOU/edit?usp=sharing)](https://docs.google.com/spreadsheets/d/1jIyvXOCuHDY8P548oaWMtbU2hrLgxuKEvgL8D92dNOU/edit?usp=sharing)
- ✅ **Final model weights**: Metrics match exactly between `main` and this PR branch
- ⚠️ **Checkpoint weights**: Discrepancies observed when comparing newer branches to `main` (tracked in #168)
- ✅ **Inference validation**: Successfully validated PSF model inference functionality

Visual Validation: The image below shows prediction quality with observation, model prediction, and residual comparison:
Screenshot 2025-07-10 at 17 31 28

## Next Steps

Further improvements to PSF inference will be implemented in the target inference branch to avoid additional scope creep in this PR.


src/wf_psf/psf_models/models/psf_model_physical_polychromatic.py
```diff
@@ -109,203 +110,183 @@ def __init__(self, model_params, training_params, data, coeff_mat=None):
...
-        data: DataConfigHandler object
-            A DataConfigHandler object providing access to training and tests datasets, as well as prior knowledge like Zernike coefficients.
-        """
-        self.zks_prior = get_zernike_prior(model_params, data, data.batch_size)
-        self.n_zks_total = max(
-            model_params.param_hparams.n_zernikes,
-            tf.cast(tf.shape(self.zks_prior)[1], tf.int32),
+        # Compute contributions once eagerly (outside graph)
+        zks_total_contribution_np = self._assemble_zernike_contributions().numpy()
+        self._zks_total_contribution = tf.convert_to_tensor(zks_total_contribution_np, dtype=tf.float32)
```
jeipollack, line 137: Check if it necessary to convert to numpy.

jeipollack:
I've validated that the refactored Physical Polychromatic PSF Field model reproduces exactly the results as the `develop` branch. I am going to proceed with merging the pull request into the inference feature branch.

PR #185, opened 2025-11-19
Feature/81/eliminate repodir dependency

### LINKED ISSUE #81: Adjust UI to remove arg --repodir
```
-h, --help            show this help message and exit
  --conffile CONFFILE, -c CONFFILE
                        a configuration file containing program settings.
  --repodir REPODIR, -r REPODIR
```

During PR, 
@[sfarrens](https://github.com/sfarrens) sfarrens [last week](https://github.com/CosmoStat/wf-psf/issues/new#discussion_r1377393988)
Why do we need to specify this?

While the code can be called from anywhere, the dataset files and SEDs are in sub-directories inside the repo dir. I guess it could be specified in the config file. I will think about it some more.

The batch job script executes the command cd $WORK/wf-psf before running WaveDiff to put us at the directory level. For the datasets, I see only the relative path is specified. It does appear that the repodir arg is only used to retrieve the logging config file.

jeipollack:
### Description
This PR removes the `--repodir` argument and updates` _setup_logging()` to load `logging.conf` from package-internal resources, simplifying the CLI and ensuring logging works reliably in HPC and batch jobs.

### Problem

Previously, WaveDiff required the `--repodir` argument only to locate `logging.conf` in the repository. This approach places an unnecessary requirement on users.

### Solution

- Removed `--repodir` references from CLI and code.
- Moved `logging.conf` into src/wf_psf/config/ as package data.
- `_setup_logging()` now uses `importlib.resources.path` with module-level import.
- Added optional fallback for development environments.
- Updated pytest tests to mock `resources.path` and `logging.config.fileConfig`.

### Testing

- Unit tests verify correct resource lookup, logfile path, and logging configuration.
- Existing directory setup methods remain functional.
- Successful code execution without `--repodir`.

### Related Issues
Fixes #81

### DIFF
--- a/src/wf_psf/tests/test_utils/conftest.py
+++ b/src/wf_psf/tests/test_utils/conftest.py
@@ -10,7 +10,6 @@ wf_psf utils package.
 
 import pytest
 import os
-from wf_psf.utils.read_config import RecursiveNamespace
 from wf_psf.utils.io import FileIOHandler
 
 cwd = os.getcwd()
@@ -24,6 +23,11 @@ def path_to_repo_dir():
 @pytest.fixture
 def path_to_tmp_output_dir(tmp_path):
     return tmp_path
+
+
+@pytest.fixture
+def path_to_test_dir(path_to_repo_dir):
+    return os.path.join(path_to_repo_dir, "src", "wf_psf", "tests")
 
 
 @pytest.fixture
--- a/src/wf_psf/tests/test_utils/conftest.py
+++ b/src/wf_psf/tests/test_utils/conftest.py
@@ -22,11 +22,6 @@ def path_to_repo_dir():
 
 
 @pytest.fixture
-def path_to_test_dir(path_to_repo_dir):
-    return os.path.join(path_to_repo_dir, "src", "wf_psf", "tests")
-
-
-@pytest.fixture
 def path_to_tmp_output_dir(tmp_path):
     return tmp_path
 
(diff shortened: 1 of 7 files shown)


jeipollack:
Hi @sfarrens, this PR is ready for review and is meant to be part of release 3.0

src/wf_psf/tests/test_utils/conftest.py
```diff
@@ -22,13 +21,13 @@ def path_to_repo_dir():
...
-    return os.path.join(path_to_repo_dir, "src", "wf_psf", "tests")
+def path_to_tmp_output_dir(tmp_path):
+    return tmp_path
 
 
 @pytest.fixture
-def path_to_tmp_output_dir(tmp_path):
-    return tmp_path
+def path_to_test_dir(path_to_repo_dir):
+    return os.path.join(path_to_repo_dir, "src", "wf_psf", "tests")
```
ChaitanyaChawak: Hi @jeipollack, shouldn't this be removed?

jeipollack: Thanks for the question. It could be removed but I need to update the fixture : `path_to_config_dir` after it that requires the absolute path to the `data` directory inside the tests package.

jeipollack: I will let you know when I pushed a commit

jeipollack: @ChaitanyaChawak  pushed the commit. I had some doubts as to whether that other fixture is really needed or not, but it would be out-of-scope. :-)

Issue #187, opened 2025-11-25
Move the`ground_truth_model` conditional out of `compute_poly_metric`


jeipollack:
### Description:
Currently, `compute_poly_metric` (and potentially other metric functions in `metrics.py`) contains conditional logic to handle the presence or absence of `ground_truth_model`.  Even if `ground_truth_model` is "present" in the dataset, the ground truth model class is instantiated anyways, and repeatedly in each `evaluate_<name_of_metric>` call. This mixes configuration handling with metric computation, causes redundancy, and makes the code harder to maintain and test.

### Proposed Options:

Move the conditional check out of `compute_poly_metric` (and other metrics) into one of the following:

1. `MetricsParamsHandler.evaluate_metrics_polychromatic_lowres`
2. A pre-validation step during `MetricsParamsHandler.__init__`, or a dedicated helper method like `prepare_ground_truth_model`

### Preferred Approach:
Option 2 – handle the ground truth model setup in the initialization/pre-validation phase. This ensures that all metrics functions can safely assume a valid `ground_truth_model` and simplifies downstream logic.

### Notes:

- This change would not alter functionality; it’s a refactor to separate configuration validation from metric computation.
- Existing tests (or new unit tests like `test_missing_ground_truth_model_raises` in #186) can confirm that missing `ground_truth_model` raises an appropriate error.

PR #188, opened 2025-11-25
Add unit tests for metrics pipeline configuration behaviour

### LINKED ISSUE #186: Explore metrics_interface behavior under different evaluation flag configurations
### Motivation

While updating the documentation for the upcoming release 3.0.0, it became clear that the optional versus required settings for the metrics pipeline are not well documented. Clarifying this, in parallel with expanding the test coverage for different flag combinations, will help users avoid misconfigurations.

### Description

Currently, the `metrics_interface_test.py` unit test verifies that evaluate_model calls the appropriate metric functions based on a single fixed configuration:

- `eval_mono_metric=True`
- `eval_opd_metric=False`
- `eval_test_shape_results_dict=True`
- `eval_train_shape_results_dict=False`

We need to systematically test other configurations to ensure the correct behavior:

- Cases where `eval_opd_metric=True`
- Cases where `eval_mono_metric=False`
- Cases where both train and test shape metrics are evaluated
- Scenarios with and without `ground_truth_model` instantiation

### Acceptance Criteria:

1. Adapt `metrics_interface_test.py` to parametrize tests with different `MetricsParamsHandler` flag settings.
2. Ensure that metric evaluation functions are called according to the flags.
3. Verify correct saving behavior (`np.save`) for each combination.
4. Document any limitations or unexpected behaviors.

jeipollack:
This PR adds unit tests to verify metrics pipeline behaviour under different configuration flags and ensures required fields are enforced.

### Key Changes
- Added unit tests to check `evaluate_model` and `MetricsParamsHandler` behaviour for different metrics config settings.
- Verified that flags correctly trigger or skip metric evaluation functions.
- Confirms that required fields (`model_params`) must be provided.
- Observed `ground_truth_model` is always instantiated even if present in dataset (opened dedicated #187 to address in a future PR/release)
- GPU-intensive integration tests remain skipped in CI, can be run locally.
- Closes #186

### DIFF
--- a/src/wf_psf/tests/test_metrics/metrics_interface_test.py
+++ b/src/wf_psf/tests/test_metrics/metrics_interface_test.py
@@ -1,8 +1,8 @@
-
 from unittest.mock import patch, MagicMock
 import pytest
 from wf_psf.metrics.metrics_interface import evaluate_model, MetricsParamsHandler
 from wf_psf.data.training_preprocessing import DataHandler
+
 
 @pytest.fixture
 def mock_metrics_params():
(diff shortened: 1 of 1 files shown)


src/wf_psf/tests/test_metrics/metrics_interface_test.py
```diff
@@ -41,34 +40,173 @@ def mock_data():
...
+        ),
+    )
+
+    with (
+        patch(
+            "wf_psf.metrics.metrics_interface.MetricsParamsHandler.evaluate_metrics_polychromatic_lowres"
+        ) as mock_poly,
+        patch(
+            "wf_psf.metrics.metrics_interface.MetricsParamsHandler.evaluate_metrics_polychromatic_lowres"
+        ) as mock_poly,
```
ChaitanyaChawak, line 161: duplicate?

jeipollack: Thanks! It is

jeipollack: Removed it.

Issue #240, opened 2026-08-03
[ARCHITECTURE] Implement Quality Control Pipeline Orchestration


jeipollack:
### Description

The Quality Control framework now provides registries for quality metrics and rejection policies. A pipeline orchestration layer is required to connect these components and provide a configurable execution flow.

The orchestration layer should:

* load enabled quality metrics from configuration;
* instantiate metric implementations through the metrics registry;
* compute quality metric values for dataset samples;
* apply configured rejection policies through the rejection policy registry;
* combine resulting validity masks for downstream filtering.

This will decouple pipeline execution logic from individual metric and rejection policy implementations.

### Scope

* Add quality control pipeline coordinator/executor.
* Define interaction between:

  * `QualityControlConfig`
  * `MetricsRegistry`
  * `RejectionPolicyRegistry`
  * quality metric implementations
  * rejection policy implementations
* Add unit tests for pipeline orchestration.

### Out of scope

* Implementation of individual metric algorithms.
* Implementation of model-dependent metrics (e.g. goodness-of-fit).
* Final optimisation of pipeline execution.

### Parent issue

Part of #226: Introduce configurable outlier analysis and rejection pipelines

jeipollack:
In working out the design issues, I came to realisation of this requirement:

> The QC pipeline must support configurable execution contexts for quality metrics. Different metrics may require different derived products (e.g. PSF predictions generated with different inference configurations). Resource generation should be centralized in the QC pipeline and shared between metrics requiring the same configuration.

jeipollack:
Here's a proposal after some back and forth with Chat:

---

### Proposed processing flow

The `QualityControlPipeline` acts as the orchestration layer responsible for evaluating dataset sample quality.

The proposed execution flow is:

```text
QualityControlPipeline.__init__(qc_config)
│
├── Load and validate QualityControlConfig
├── Build MetricsRegistry
└── Build RejectionPolicyRegistry


QualityControlPipeline.run(dataset)
│
├── Resolve enabled quality metrics
├── Determine required execution resources
├── Build QualityControlContext
│     ├── dataset
│     └── derived resources (e.g. PSF predictions - "standard" or "oversampled")
├── Instantiate enabled metric objects
├── Compute quality metrics
├── Apply rejection policies
└── Return QualityControlResult
```

---

### Design rationale

The goal is to keep responsibilities clearly separated.

* **Quality metrics** compute numerical measures describing dataset sample quality.
[...]

Issue #226, opened 2026-07-16
[ARCHITECTURE] Introduce configurable outlier analysis and rejection pipelines


jeipollack:
Define the framework for implementing reusable and extensible data quality analysis and rejection pipelines in `wf-psf`.

The pipeline should support multiple independent data quality metrics and rejection policies while maintaining a single, consistent mechanism for filtering aligned WaveDiff datasets during data preparation.

A key design principle is to **decouple metric computation from rejection policy**, allowing quality metrics to be reused for diagnostics, validation, and monitoring independently of threshold-based filtering.

```
quality metric
      ↓
metric value
      ↓
rejection policy
      ↓
boolean mask
```
Multiple rejection masks are combined into a single validity mask, which is applied to the aligned dataset.

```
Configuration
      |
      v
Quality Metric Computation
      |
      +--> mask obscuration
      |
      +--> goodness-of-fit
      |
      v
Quality Metrics
      |
      v
Rejection Policies
      |
      v
Combined validity mask
      |
      v
Dataset filtering
```

The framework should distinguish between:
- **Data validity checks** required to construct a usable dataset (e.g. finite-value checks during data preparation).
- **Scientific quality criteria** used to assess or reject samples based on configurable policies

## Goals

- [ ]  Introduce a configurable quality analysis and rejection pipeline.
- [ ]  Support multiple independent data quality metrics.
- [ ]  Allow metrics to be computed independently from whether they are used for rejection.
- [ ]  Produce a single combined validity mask from configured rejection policies.
- [ ]  Preserve sample alignment when filtering WaveDiff datasets.
- [ ]  Allow future quality metrics and rejection methods to be added without modifying the pipeline architecture.
- [ ]  Keep the framework agnostic to specific quality metrics and rejection strategies.

jeipollack:
it terms of package structure I am converging on :

```
quality_control/

    config.py              # QualityControlConfigHandler

    pipeline.py            # orchestrator

    reporting.py

    metrics/
        base.py
        registry.py
        mask_obscuration.py
        goodness_of_fit.py

    rejection/
        base.py
        registry.py
        threshold.py
```

jeipollack:
For rejection policies in `rejection`, we can extend it cleanly with other types of policies, e.g.:

- `PercentilePolicy`
- `SigmaClipPolicy`
- `TopNPolicy`

PR #242, opened 2026-08-09
241 architecture extend quality control configuration

### LINKED ISSUE #241: [ARCHITECTURE] Extend Quality Control Configuration for Resource Dependencies
## Summary

Extend the Quality Control configuration framework to support declaring resources available to the QC pipeline and specifying the resources required by individual quality metrics.

The configuration should provide a generic mechanism for describing resources without coupling the configuration layer to specific resource types such as PSF models.

This configuration will provide the basis for the `QualityControlPipeline` resource-resolution and execution-context architecture defined in #240.

## Proposed configuration

```yaml
resources:

  psf_models:
    standard:
      inference_config: inference_standard.yaml

metrics:

  mask_obscuration:
    enabled: true
    params:
      aperture: gaussian
      sigma: 2.5

  goodness_of_fit:
    enabled: true
    required_resources:
      - psf_models.standard
    params:
      statistic: reduced_chi_square
      normalize_residuals: true

rejection:

  mask_obscuration:
    enabled: true
    threshold: 0.25

  goodness_of_fit:
    enabled: false
    threshold: 3.0

reporting:

  save_metrics: true
  log_statistics: true
```

The `resources` section is intentionally generic so that future resource types can be introduced without modifying the configuration framework.

## What’s changed

* Add a `ResourcesConfig` configuration object for resources available to the QC pipeline.
* Add `required_resources` to `QualityMetricConfig`.
* Preserve `params` for metric-specific configuration.
* Parse the new `resources` section.
* Parse metric `required_resources` declarations.
* Validate that required resource identifiers follow the expected `<resource_type>.<resource_name>` form.
* Validate that required resources are available in the configuration.
* Validate that enabled rejection policies reference configured and enabled quality metrics.
* Run cross-section configuration validation when loading the QC configuration.
* Update configuration examples and unit tests accordingly.
* Add a changelog fragment under **Internal changes**.

## Validation

The configuration layer should reject, with clear errors:

* malformed resource identifiers;
* unknown resource types;
* unknown resource names;
* enabled rejection policies referring to unknown metrics;
* enabled rejection policies referring to disabled metrics.

Validation should occur during `QualityControlConfigHandler.load()`, so callers receive either a valid `QualityControlConfig` or an exception.

## Relationship to #240

This issue provides the configuration plumbing required by the QC pipeline orchestration implemented in #240.

The pipeline will use `required_resources` to determine which execution resources need to be prepared before metric evaluation and will subsequently make those resources available through the QC execution context.

Resource generation and metric execution are **not** part of this issue.

## Scope / Non-goals

This issue does not implement:

* PSF inference or other resource generation;
* `QualityControlContext`;
* resource deduplication or caching;
* quality metric computation;
* rejection policy execution;
* shape metrics;
* pipeline execution orchestration.

Those responsibilities remain part of the QC pipeline implementation and subsequent metric-specific work.

## Acceptance criteria

* [x] `ResourcesConfig` represents configured resources.
* [x] `QualityMetricConfig` supports `required_resources`.
* [x] Resource and metric configuration are parsed correctly.
* [x] Cross-section configuration validation is performed during configuration loading.
* [x] Invalid resource references produce clear `ValueError`s.
* [x] Invalid rejection-policy/metric references produce clear `ValueError`s.
* [x] Existing configuration tests continue to pass.
* [x] New unit and integration tests cover the validation behavior.
* [x] Working QC configuration example is updated to contain only currently supported functionality.
* [x] Changelog fragment is added under **Internal changes**.

jeipollack:
## Summary

Extends the Quality Control configuration framework to support configurable execution resources and metric resource requirements.

The configuration now allows quality metrics to declare the resources they require, while the configuration loader validates that those resources are available and that enabled rejection policies reference enabled quality metrics.

This provides the configuration foundation required by the Quality Control pipeline orchestration.

closes #241

## What’s changed

* Added configurable resource declarations through `ResourcesConfig`.
* Added `required_resources` to `QualityMetricConfig` so metrics can declare the resources required for their computation.
* Extended the QC YAML configuration with a generic `resources` section
* Moved `inference_config` from metric-specific `params` to the `resources` section.
* Updated the example QC configuration with more realistic metric-specific `params` arguments and values, while preserving `params` as the mechanism for metric-specific configuration.
* Added validation of metric resource identifiers and availability.
* Added validation to ensure enabled rejection policies reference configured and enabled quality metrics.
* Integrated cross-section configuration validation into `QualityControlConfigHandler.load()`.
* Updated configuration examples and tests to cover resource declarations and validation.

## How to test / verify

* Run the Quality Control configuration unit tests:
  `pytest tests/quality_control/test_config.py`
* Verify that valid QC configurations load successfully and invalid resource or rejection-policy references raise the expected `ValueError`.
* Run the full test suite to verify that existing configuration behavior remains compatible.

## Scope

* [ ] Feature
* [ ] Bug fix
* [ ] Hotfix
* [ ] Documentation / process change
* [x] Internal / refactor
* [ ] Release

This PR is part of the Quality Control architecture introduced under #226 and provides configuration plumbing required by #240 (Quality Control pipeline orchestration).

## Changelog

* [x] Changelog fragment added (if applicable)

## Reviewer Checklist

* [x] The PR targets the correct base branch (`develop`, or `main` for release PRs)
* [x] The PR is assigned to the developer
* [x] Appropriate labels are applied
* [x] The PR is included in relevant projects and/or milestones
* [x] Description clearly explains what has changed
* [x] Issue references included, if applicable
* [x] Code and documentation adhere to current standards (`ruff`)
* [x] Documentation updates included, if relevant
* [x] CI tests are passing
* [x] All reviewer comments have been addressed

## Next Steps / Notes

The resource configuration is intentionally generic. The current implementation supports declaring resources such as standard PSF inference configurations, while the pipeline orchestration will resolve and prepare the resources required by enabled metrics.

Resource generation, execution-context construction, and metric computation remain part of the Quality Control pipeline implementation in #240 and subsequent metric-specific work.


src/wf_psf/quality_control/config.py
```diff
@@ -143,10 +240,127 @@ def parse_reporting_config(config: dict | None) -> ReportingConfig:
...
+    ResourcesConfig
+        Parsed resource configuration.
+
+    Raises
+    ------
+    TypeError
+        If the configuration section is not a mapping.
+    """
+    if config is None:
+        return ResourcesConfig()
```
roryclaydon1994, line 271: How is the empty config likely to come about in production and what function does it serve? I'm curious if the intent could also be handled by just an empty dictionary or indeed if it needs to be handled at all?

In terms of this factory method, it might more naturally fit as a classmethod to keep it associated with the class? Then it would also be inheritable if that's of use.

src/wf_psf/quality_control/config.py
```diff
@@ -143,10 +240,127 @@ def parse_reporting_config(config: dict | None) -> ReportingConfig:
...
+
+    if not isinstance(config, dict):
+        raise TypeError("Resources configuration must be a mapping.")
```
roryclaydon1994: Not super important, but this is probably more restrictive than required. Anything appropriate getter/setters would work I think?

src/wf_psf/quality_control/config.py
```diff
@@ -143,10 +240,127 @@ def parse_reporting_config(config: dict | None) -> ReportingConfig:
...
+
+            resources = config.resources.available
+
+            if (
+                resource_type not in resources
+                or resource_name not in resources[resource_type]
+            ):
+                raise ValueError(
+                    f"Metric '{metric_name}' requires unknown resource '{req}'."
+                )
```
roryclaydon1994, line 329: Is there a way to simplify the logic here?

E.g. a try/except block? 

Could probably also add a check that there is only one "." in case of a typo

src/wf_psf/quality_control/config.py
```diff
@@ -143,10 +240,127 @@ def parse_reporting_config(config: dict | None) -> ReportingConfig:
...
+        Parsed quality control configuration.
+
+    Raises
+    ------
+    ValueError
+        If an enabled rejection policy references an unknown or disabled
+        quality metric.
+    """
+    for metric_name, metric_rejection_policy in config.rejection.items():
+        if metric_rejection_policy.enabled:
```
roryclaydon1994: Just a style thing: the amount of indentation could be reduced e.g.

```
if not metric_rejection_policy.enabled:
    continue

# Same block follows but one level of indentation in
```

not required for PR, just a personal preference.

src/wf_psf/tests/test_quality_control/config_test.py
```diff
@@ -22,16 +29,70 @@ def load_config(config_file: str) -> QualityControlConfig:
...
+        rejection=None,
+    ):
+        return QualityControlConfig(
+            metrics=metrics
+            or {
+                "goodness_of_fit": QualityMetricConfig(
+                    enabled=True,
+                    required_resources=required_resources or [],
+                )
+            },
```
roryclaydon1994: To make this section easier to parse for a reader, perhaps the default options could be separately defined
```
        metric_default={
                "goodness_of_fit": QualityMetricConfig(
                    enabled=True,
                    required_resources=required_resources or [],
                )
            }
        return QualityControlConfig(
            metrics=metrics  or  metric_default,
           ...
```

roryclaydon1994 (commented):
## Review Feedback

Great job, architecture clean and introduces minimal coupling and tests are comprehensive. Please see inline comments.

src/wf_psf/quality_control/config.py
jeipollack: Thanks for these questions!

An empty resources configuration is expected when the enabled metrics do not require any additional derived resources. For example, `mask_obscuration` can operate on data already provided by the calling pipeline task, so a configuration containing only this metric would legitimately have no additional resources.

I think returning an empty `ResourcesConfig` gives the configuration a consistent typed interface (`config.resources.available`) rather than requiring callers to handle `None` or a plain dictionary when no resources are configured.

Regarding the classmethod suggestion, I agree that `parse_resources_config` could naturally be associated with `ResourcesConfig`. However, the section parsers currently form a common parsing layer used through `SECTION_PARSERS`, and I think keeping them as module-level functions maintains a consistent separation between YAML parsing and the configuration dataclasses. I would therefore prefer to keep the current approach unless there is a specific benefit to making the parsers classmethods. I am open to revisit this if we think configuration parsing belongs more naturally on the individual configuration classes.

src/wf_psf/quality_control/config.py
jeipollack: I generalised the accepted input type by replacing `dict` with `Mapping`, so other mapping implementations are supported as well.

src/wf_psf/quality_control/config.py
jeipollack: I agree that the identifier validation can be simplified. I don't think a try/except provides much benefit here since we also need to check for empty components, but I've changed the parsing to split on "." and explicitly require exactly two non-empty components. This also catches identifiers containing multiple "." characters, such as `psf_models.standard.foo`. I also consolidated the unit tests for malformed and unknown resource identifiers using `pytest.mark.parametrize`.

src/wf_psf/quality_control/config.py
jeipollack: Thanks, I agree. it's tidier now.

src/wf_psf/tests/test_quality_control/config_test.py
jeipollack: Thanks, I updated the fixture with default options.

jeipollack:
Thanks, @roryclaydon1994, again for the careful and thoughtful review! This iteration helped me identify a few additional improvements beyond the specific comments, including adding an explicit type check for `required_resources` with corresponding tests. I also found an unused/deprecated YAML fixture while reviewing the test configuration files and removed it.
I've addressed the review comments and pushed the updates. Ready for another look when you have a chance.

PR #243, opened 2026-08-18
240 architecture implement quality control pipeline orchestration

### LINKED ISSUE #240: [ARCHITECTURE] Implement Quality Control Pipeline Orchestration

jeipollack:
## Summary

This PR implement a minimal quality control pipeline orchestrator; it builds the processing sequence for resource resolution, metric computation, and policy application.

Closes #240 #246 

## What’s changed

- Add `QualityControlContext` class definition
- Define `QualityControlResult` for metric results and rejection masks
- Add helpers to instantiate enabled metrics and rejection policies
- Add value attribute to `ThresholdRejectionPolicy`
- Build the processing sequence for resource resolution, metric computation, and policy application
- Add unit and integration tests for the orchestration

## How to test / verify

- Check CI tests pass.
- Confirm the integration test covers resource resolution, metric computation, and rejection-mask generation.

## Scope

> Indicate the type of PR:

- [ ] Feature  
- [ ] Bug fix  
- [ ] Hotfix  
- [X] Internal / refactor  
- [ ] Release

## Changelog

The quality control pipeline interface and configuration are still being finalised. A changelog fragment will be added with the PR that finalises the user-facing pipeline behaviour.

## Reviewer Checklist

- [x] CI tests are passing

The next step is to implement preparation of missing resources so that the resource resolution stage can provide all resources required by enabled quality metrics.

### DIFF
--- a/src/wf_psf/quality_control/context.py
+++ b/src/wf_psf/quality_control/context.py
@@ -19,7 +19,7 @@ class QualityControlContext:
 
     Attributes
     ----------
-    dataset : dict[str, Any]
+    dataset : Any
         Dataset or data container supplied to the quality control pipeline.
 
     resources : dict[str, Any]
@@ -27,5 +27,5 @@ class QualityControlContext:
         resource identifier.
     """
 
-    dataset: Any = None
+    dataset: Any
     resources: dict[str, Any] = field(default_factory=dict)
--- a/src/wf_psf/quality_control/context.py
+++ b/src/wf_psf/quality_control/context.py
@@ -1,0 +1,31 @@
+"""Quality control context.
+
+Encapsulates contextual information, such as datasets and resolved resources,
+required by the quality control pipeline and its metrics.
+
+:Authors:
+    Jennifer Pollack <jennifer.pollack@cea.fr>
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+
+@dataclass
+class QualityControlContext:
+    """Context shared across the quality control pipeline.
+
+    Attributes
+    ----------
+    dataset : dict[str, Any]
+        Dataset or data container supplied to the quality control pipeline.
+
+    resources : dict[str, Any]
+        Ready-to-use resources required by enabled quality metrics, keyed by
+        resource identifier.
+    """
+
+    dataset: Any = None
+    resources: dict[str, Any] = field(default_factory=dict)
--- a/src/wf_psf/quality_control/pipeline.py
+++ b/src/wf_psf/quality_control/pipeline.py
@@ -2,11 +2,10 @@
 
 Defines the orchestration layer for dataset quality control.
 
-The QualityControlPipeline coordinates quality metric evaluation,
-application of rejection policies, reporting, and dataset filtering.
-Individual quality metrics and rejection policies are provided through
-their respective interfaces, allowing new methods to be added without
-modifying the pipeline implementation.
+The QualityControlPipeline coordinates quality metric evaluation and
+sample rejection. Individual quality metrics and rejection policies are
+provided through their respective interfaces, allowing new methods to
+be added without modifying the pipeline implementation.
 
 :Authors: Jennifer Pollack <jennifer.pollack@cea.fr>
 
@@ -14,6 +13,12 @@ modifying the pipeline implementation.
 
 from dataclasses import dataclass
 import numpy as np
+
+from wf_psf.quality_control.config import QualityControlConfigHandler
+from wf_psf.quality_control.metrics.base import QualityMetric
+from wf_psf.quality_control.metrics.registry import build_metrics_registry
+from wf_psf.quality_control.rejection.registry import build_rejection_policy_registry
+
 
 @dataclass
 class QualityControlResult:
@@ -50,15 +55,27 @@ class QualityControlPipeline:
     of the pipeline execution.
     """
 
-    def __init__(
-            self,
-            metrics,
-            rejection_policies,
-    ):
-        ...
+    def __init__(self, qc_config_path):
+        self.config = QualityControlConfigHandler(qc_config_path).load()
+        self.metrics_registry = build_metrics_registry()
+        self.rejection_registry = build_rejection_policy_registry()
+
+    def _instantiate_metrics(self) -> dict[str, QualityMetric]:
+        """Instantiate enabled quality metric implementation from configuration."""
+        metrics = {}
+
+        for name, metric_config in self.config.metrics.items():
+            if not metric_config.enabled:
+                continue
+
+            metric_cls = self.metrics_registry.get(name)
+
+            metrics[name] = metric_cls()
+
+        return metrics
 
     def run(self, dataset):
-
+        """Run quality control pipeline."""
+        #        metrics = self._instantiate_metrics()
         ...
-
         return QualityControlResult(...)
--- a/src/wf_psf/quality_control/pipeline.py
+++ b/src/wf_psf/quality_control/pipeline.py
@@ -22,6 +22,10 @@ from wf_psf.quality_control.rejection.base import RejectionPolicy
 from wf_psf.quality_control.rejection.registry import build_rejection_policy_registry
 from wf_psf.quality_control.resources import Resources
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 @dataclass
 class QualityControlResult:
@@ -80,11 +84,14 @@ class QualityControlPipeline:
 
         for name, metric_config in self.config.metrics.items():
             if not metric_config.enabled:
+                logger.debug("Skipping metric %s: not enabled.", name)
                 continue
 
             metric_cls = self.metrics_registry.get(name)
 
             metrics[name] = metric_cls()
+
+        logger.debug("Instantiated metrics: %s", list(metrics))
 
         return metrics
 
@@ -105,12 +112,16 @@ class QualityControlPipeline:
 
         for metric_name, rejection_config in self.config.rejection.items():
             if not rejection_config.enabled:
+                logger.debug("Skipping rejection policy %s: not enabled.", metric_name)
                 continue
 
             policy_name, policy_params = next(iter(rejection_config.policy.items()))
             policy_cls = self.rejection_registry.get(policy_name)
 
             rejection_policies[metric_name] = policy_cls(**policy_params)
+
+        logger.debug("Instantiated rejection policies: %s", list(rejection_policies))
+
         return rejection_policies
 
     def run(self, dataset, provided_resources=None):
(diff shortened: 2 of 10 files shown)


src/wf_psf/quality_control/context.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclass
+class QualityControlContext:
+    """Context shared across the quality control pipeline.
+
+    Attributes
+    ----------
+    dataset : Any
+        Dataset or data container supplied to the quality control pipeline.
```
roryclaydon1994, line 23: I was imagining datasets as a list of `np.arrays`. What else could they be?

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
+        Notes
+        -----
+        The quality control configuration is assumed to have been validated
+        before policy instantiation.
+        """
+        metrics = {}
+
+        for name, metric_config in self.config.metrics.items():
+            if not metric_config.enabled:
+                continue
```
roryclaydon1994, line 88: Perhaps some debug logging statements could be useful in this function or a final info log statement as to what was loaded?

Same comment for the other `_instantiate_` function

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
+
+        provided_resources : Mapping[str, Any] or None
+            Ready-to-use resources supplied by the pipeline caller, keyed by resource identifier.
+
+        Notes
+        -----
+        The pipeline is expected to be invoked only when at least one quality metric is enabled in the quality control configuration.
+        """
+        resource_manager = Resources(self.config)
+        resolved_resources = resource_manager.resolve(provided_resources)
```
roryclaydon1994: Perhaps this could also be moved to `_instantiate_resources` to match the other methods which query the config?

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
+
+        context = QualityControlContext(dataset, resolved_resources)
+
+        metrics = self._instantiate_metrics()
+
+        metric_results = {
+            name: metric.compute(context) for name, metric in metrics.items()
+        }
```
roryclaydon1994, line 169: I'm assuming this could be quite a compute heavy loop?

If so, I think extra logging could be useful here. I guess the `compute` method itself will also have some logging or `tqdm` type output?

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
+
+        rejection_policies = self._instantiate_rejection_policies()
 
-    def run(self, dataset):
+        rejection_masks = {
+            name: policy.apply(metric_results[name])
+            for name, policy in rejection_policies.items()
+        }
```
roryclaydon1994: I guess this generally is quite light compare to the metric computation?

At any point in the future, would metric computation and rejection be performed in one pass or are these both so inexpensive it doesn't matter? My angle is primarily from the GPU perspective of minimising memory traffic / fusing operations if possible

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
 
-        ...
+        if rejection_masks:
+            valid_mask = np.logical_and.reduce(list(rejection_masks.values()))
```
roryclaydon1994: Please add the comment that this means that the reduction here means that a valid mask is where for each element of a dataset, all the rejection masks were true.

Just to check, `valid_mask` means the ones to use right? But if `rejection_mask` is `True`, what does that mean? Should it be rejected?

src/wf_psf/tests/test_quality_control/pipeline_test.py
```diff
@@ -0,0 +1,123 @@
...
+    assert isinstance(rejection_policies["mask_obscuration"], ThresholdRejectionPolicy)
+    assert rejection_policies["mask_obscuration"].value == 3.0
+
+    assert "goodness_of_fit" not in rejection_policies
+
+
+# Test pipeline runner
+def test_pipeline_run(pipeline_factory):
+    metric_result = np.array([1.0, 2.0, 3.0])
+    rejection_mask = np.array([True, False, True])
```
roryclaydon1994: Should the rejection mask be a list with an entry per metric?
So here I guess like
```
rejection_mask = np.array(
    [True, False, True],
    [True, False, False]
)
```
Then you could check the `np.logical_and.reduce` works as expected?

src/wf_psf/quality_control/pipeline.py
```diff
@@ -50,15 +58,102 @@ class QualityControlPipeline:
...
-            metrics,
-            rejection_policies,
-    ):
-        ...
+    def __init__(self, qc_config_path):
+        self.config = QualityControlConfigHandler(qc_config_path).load()
+        self.metrics_registry = build_metrics_registry()
+        self.rejection_registry = build_rejection_policy_registry()
+
+    def _instantiate_metrics(self) -> dict[str, QualityMetric]:
```
roryclaydon1994, line 70: Private so the API can be changed without major releases or for security?

roryclaydon1994 (commented):
## Review Feedback

Very complete, great tests and design.

I think in terms of design, I would suggest splitting out the instantiate components of the `run` method of `QualityControlPipeline` so that run only has the responsibility of running and all loading and resolving responsibilities are in a separate method.

I would also suggest sprinkling in some logging statements as you see fit to help observability.

src/wf_psf/quality_control/context.py
jeipollack: I was wondering about this and settled on using `Any` here because I didn't think the orchestrator should assume a particular dataset representation, such as a list of arrays with a fixed positional meaning. If we use a list, we'd also need to define what each array represents and enforce that ordering.

The intent is for the dataset/data container to be passed through the `QualityControlContext` to the metric implementations, which can interpret the data according to their requirements. I agree that if different metrics require incompatible representations, there would need to be some mapping or adapter somewhere; I just don't think that mapping belongs in the orchestrator itself. For example, a user-defined metric could adapt the canonical dataset representation to its own requirements.

If we want to define a specific canonical dataset contract, I think that is a separate API design decision and could be addressed in a subsequent PR rather than something the orchestrator should assume here.

src/wf_psf/quality_control/pipeline.py
jeipollack: thanks! done.

src/wf_psf/quality_control/pipeline.py
jeipollack: I've added a `_resolve_resources()` helper to keep the resource handling consistent with the other pipeline helpers. It currently delegates to `Resources.resolve()` and will be extended to pass the dataset when resource preparation is implemented.

src/wf_psf/quality_control/pipeline.py
jeipollack: Yes, I expect `compute()` could be compute-heavy depending on the metric implementation. I prefer to delegate detailed logging or progress reporting to the metric's `compute()` method, since the type and granularity of the messages may differ between metrics.

The orchestrator could log which metrics are being computed if that would be useful, but I don't think it should duplicate metric-specific progress or performance logging here. Lmk what you prefer. :-)

src/wf_psf/quality_control/pipeline.py
jeipollack: Could we investigate in a future optimisation PR? With the current design, `policy.apply()` is intended to be a lightweight operation that takes an already-computed metric and produces a validity mask according to the configured policy, so I don't think it needs to run on the GPU independently.

Does the validation pipeline currently run on the GPU, or is GPU execution specific to some of the metric computations ?

At this dev stage, I would prefer to keep metric computation and rejection as separate concerns in the API rather than combining them just to enable fusion. If profiling later shows that memory traffic between metric computation and rejection is significant, we could investigate a fused implementation or a more GPU-aware execution path without changing the current conceptual separation.

src/wf_psf/quality_control/pipeline.py
jeipollack: Sorry for the confusion. I clarified the semantics by renaming `rejection_masks` to `validity_masks`, updating the `RejectionPolicy` ABC `apply()` method docstring to explicitly define the boolean mask, and adding the same clarification as a comment where the validity masks are combined.

src/wf_psf/tests/test_quality_control/pipeline_test.py
jeipollack: Not a list, but a dict with the `metric_name` as the key and the corresponding `validity_mask` as the value. I added a new test, `test_pipeline_run_multiple_rejection_policies`, to verify that the individual validity masks are combined correctly using `np.logical_and.reduce()`.

src/wf_psf/quality_control/pipeline.py
jeipollack: Honestly, I marked these methods as private because they are internal orchestration helpers rather than part of the public API of the pipeline. The public entry point is run(), while `_instantiate_metrics()`, `_instantiate_rejection_policies()`, and now `_resolve_resources()` are implementation details used by the pipeline to build the processing sequence.

My understanding is that the leading underscore signals that these methods are not intended to be called directly by users, and therefore gives us some freedom to change their implementation without treating those changes as public API changes.

jeipollack:
Thanks a mill, @roryclaydon1994! I addressed each comment and made some updates. I also corrected some syntax errors in the public `quality_control.yaml` file which I discovered during my dev meeting with my colleagues.

PR #245, opened 2026-08-20
244 architecture resolve resource dependencies

### LINKED ISSUE #244: [ARCHITECTURE] Resolve quality control resource dependencies
### Objective

To provide flexibility in how resources required by quality metrics are obtained, the quality control pipeline should support both **caller-provided resources** and resources that need to be generated or otherwise prepared by the QC pipeline.

The QC configuration already identifies resources required by enabled metrics through `required_resources`. The pipeline therefore needs a resource module that can:

* determine which resources are required by the enabled quality metrics;
* identify which of those resources have already been supplied by the caller;
* determine which required resources are still missing and need to be generated or otherwise prepared.

For example, a caller may already have generated PSF model predictions before invoking the QC pipeline. In that case, QC should consume the supplied predictions rather than regenerate them.

### Scope

* Introduce resource dependency handling separate from the pipeline orchestration logic.
* Determine required resources from enabled quality metrics.
* Resolve required resources against resources supplied by the caller.
* Return the resolved resources and the resources that still require preparation.
* Add unit tests covering resource dependency and resolution behaviour.

### Out of scope

The mechanism for **generating or acquiring missing resources** is not defined by this issue.

In particular, this issue does not introduce a `ResourceProvider`, `ResourceResolver`, or PSF inference provider abstraction. The appropriate mechanism for preparing missing resources can be addressed once the requirements for different QC users and upstream pipelines are better understood.

### Expected flow

```text
QualityControlConfig
        │
        ▼
required resources from enabled metrics
        │
        ▼
resolve against caller-provided resources
        │
        ├── resolved resources
        │
        └── missing resources
                  │
                  ▼
       resource preparation
          (future work)
```

jeipollack:
## Summary

Adds resource dependency helpers to identify resources required by enabled
quality metrics, determine which required resources are ready to use and
which still require preparation.

Closes #244

## What’s changed

- Add `resource.py` module with a helper to determine resources required by enabled metrics
- Add a helper to identify ready-to-use resources and resources requiring preparation
- Add an orchestrator helper to resolve required resources and raise an error when resources are unavailable
- Add `resource_test.py` with tests for both helpers
- Remove resource dependency code from `quality_control.config.py` and `config_test.py`
- Move the `qc_config_factory` fixture to a new `conftest.py` in the `quality_control` test package

## How to test / verify

- Check CI passes

## Scope

> Indicate the type of PR:

- [ ] Feature
- [ ] Bug fix
- [ ] Hotfix
- [X] Internal / refactor
- [ ] Release

> This PR is part of the larger quality control pipeline orchestration
> milestone.

## Changelog

> Does this PR introduce user-visible changes? Not yet. The resource resolution functionality is an internal architectural component; user-facing behaviour will be introduced as the QC pipeline orchestration is developed.

## Reviewer Checklist

- [x] CI tests are passing

Resource preparation is intentionally not implemented in this PR. The current resource module identifies missing resources so that a subsequent stage can determine how they should be prepared or supplied.

### DIFF
--- a/src/wf_psf/quality_control/resources.py
+++ b/src/wf_psf/quality_control/resources.py
@@ -13,105 +13,66 @@ from typing import Any
 from wf_psf.quality_control.config import QualityControlConfig
 
 
-def get_required_resources(
-    config: QualityControlConfig,
-) -> set[str]:
-    """Return resources required by enabled quality metrics.
+class Resources:
+    """Manage resources required by quality control metrics.
 
-    Parameters
-    ----------
-    config : QualityControlConfig
-        Validated quality control configuration.
+    Assesses resource requirements and availability for a validated quality
+    control configuration.
+    """
 
-    Returns
-    -------
+    def __init__(self, config: QualityControlConfig):
+        self.config = config
+
+    def get_required(self) -> set[str]:
+        """Return resources required by enabled quality metrics.
+
+        Returns
+        -------
         Unique resource identifiers required by enabled quality metrics.
 
-    Notes
-    -----
-    This configuration is assumed to have been validated for internal consistency between resource requirements and available resources.
-    """
-    return {
-        resource
-        for metric in config.metrics.values()
-        if metric.enabled
-        for resource in metric.required_resources
-    }
+        Notes
+        -----
+        This configuration is assumed to have been validated for internal consistency between resource requirements and available resources.
 
+        """
+        return {
+            resource
+            for metric in self.config.metrics.values()
+            if metric.enabled
+            for resource in metric.required_resources
+        }
 
-def resolve_resources(
-    required: set[str],
-    provided: Mapping[str, Any],
-) -> tuple[dict[str, Any], set[str]]:
-    """Separate supplied resources from resources requiring preparation.
+    def resolve(
+        self,
+        provided: Mapping[str, Any] | None = None,
+    ) -> dict[str, Any]:
+        """Resolve resources required by enabled quality metrics.
 
-    Parameters
-    ----------
-    required : set[str]
-        Unique resource identifiers required by enabled quality metrics.
+        Parameters
+        ----------
+        provided : Mapping[str, Any] or None
+            Ready-to-use resources supplied by the pipeline caller.
 
-    provided : Mapping[str, Any]
-        Ready-to-use resources provided by the quality control pipeline caller, keyed by resource identifier.
+        Returns
+        -------
+        dict[str, Any]
+            Resources required by enabled quality metrics and supplied by the
+            caller.
 
-    Returns
-    -------
-    tuple[dict[str, Any], set[str]]
-        A tuple containing:
+        Raises
+        ------
+        NotImplementedError
+            If required resources are not supplied by the caller. Preparation of
+            missing resources is not yet implemented.
+        """
+        required = self.get_required()
+        provided = {} if provided is None else provided
 
-        - a dictionary of required resources that were provided, keyed by
-          resource identifier;
-        - a set of required resource identifiers that were not provided and
-          therefore require preparation.
+        missing = required - provided.keys()
 
-    Notes
-    -----
-    This function only resolves which required resources are already
-    available. It does not prepare or otherwise acquire missing resources.
-    """
-    resolved = {
-        resource: provided[resource] for resource in required if resource in provided
-    }
-    missing = required.difference(resolved)
-    return resolved, missing
+        if missing:
+            raise NotImplementedError(
+                f"Required resources are not available: {sorted(missing)}"
+            )
 
-
-def resolve_required_resources(
-    config: QualityControlConfig,
-    provided: Mapping[str, Any] | None = None,
-) -> dict[str, Any]:
-    """Resolve resources required by enabled quality metrics.
-
-    Parameters
-    ----------
-    config : QualityControlConfig
-        Quality control configuration.
-
-    provided : Mapping[str, Any] or None
-        Ready-to-use resources supplied by the pipeline caller.
-
-    Returns
-    -------
-    dict[str, Any]
-        Resources required by enabled quality metrics and supplied by the
-        caller.
-
-    Raises
-    ------
-    NotImplementedError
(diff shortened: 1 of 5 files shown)


jeipollack:
@roryclaydon1994 I am going to switch to draft mode as I am about to add a new resource orchestration helper with some tests. I will ping you when it's ready for review

jeipollack:
@roryclaydon1994 I added the new commits and updated the description. This PR is ready for review. thanks!

src/wf_psf/quality_control/resources.py
```diff
@@ -0,0 +1,117 @@
...
+    config : QualityControlConfig
+        Validated quality control configuration.
+
+    Returns
+    -------
+        Unique resource identifiers required by enabled quality metrics.
+
+    Notes
+    -----
+    This configuration is assumed to have been validated for internal consistency between resource requirements and available resources.
```
roryclaydon1994: Could perhaps add a link to where that happens so it's easy to hop to it in the docs?

src/wf_psf/tests/test_quality_control/conftest.py
```diff
@@ -0,0 +1,50 @@
...
+from wf_psf.quality_control.config import (
+    QualityControlConfig,
+    QualityMetricConfig,
+    RejectionPolicyConfig,
+    ResourcesConfig,
+)
+
+
+@pytest.fixture
+def qc_config_factory():
```
roryclaydon1994, line 11: Is this factory the same across all tests or can it be modified? I think you could probably mark it as a session level fixture (not that there is much performance difference in doing so).

src/wf_psf/quality_control/resources.py
```diff
@@ -0,0 +1,117 @@
...
+        caller.
+
+    Raises
+    ------
+    NotImplementedError
+        If required resources are not supplied by the caller. Preparation of
+        missing resources is not yet implemented.
+    """
+    required = get_required_resources(config)
+    provided = {} if provided is None else provided
```
roryclaydon1994: If provided is None, everything is missing already, no need for resolution?

src/wf_psf/quality_control/resources.py
```diff
@@ -0,0 +1,117 @@
...
+    Notes
+    -----
+    This configuration is assumed to have been validated for internal consistency between resource requirements and available resources.
+    """
+    return {
+        resource
+        for metric in config.metrics.values()
+        if metric.enabled
+        for resource in metric.required_resources
+    }
```
roryclaydon1994: Can a metric be disabled in config, but passed in as provided?

roryclaydon1994 (commented):
## Review Feedback

Tests are very complete, good documentation and typing throughout.

In terms of design, it looks like `resources.py` could be condensed to one simple function or a class with method names that are easier to distinguish and could be overloaded in future if desired.

Re the Resources class my meaning is  `get_required_resources` and `resolve_required_resources`  become `get_required`/`filter_enabled` and `resolve` as the resource class already provides the fact it is related to resources.

src/wf_psf/quality_control/resources.py
jeipollack: With the previous helper separation I see the point. After consolidating the logic into `Resources.resolve()`, I retained the normalisation so `None` represents an empty set of caller-provided resources and the method can consistently determine which required resources are missing. This also keeps the provided handling in one place. Also, the normalization is intentional to preserve the `resolution → missing-resource → preparation` flow,  which will be implemented next.

src/wf_psf/quality_control/resources.py
jeipollack: Yes, a resource can be passed in as provided even if it isn't required by any enabled metric. In the migrated code, the handling of this behaviour is covered by the `get_required()` tests in the updated implementation. The `resolve()` tests separately cover filtering the caller-provided resources against the required resources. For example, when `psf_models.standard` is required but both standard and oversampled are provided, only standard is returned as resolved.

jeipollack:
Thanks @roryclaydon1994 again for your thoughtful review. I followed the suggestion in your feedback and introduced a `Resources` class to act a resource manager and moved and consolidated the resource helper methods. I am ready for you to take a second look.

src/wf_psf/quality_control/resources.py
roryclaydon1994: I still meant to resolve as described, just skipping straight to processing all missing (same as all required in this case) resources

src/wf_psf/quality_control/resources.py
roryclaydon1994: What does it mean for a resource to be provided, but not required? Would it be used?

jeipollack: I see this primarily as allowing a caller to provide a broader set of resources than the current configuration requires, for example when setting up shared/default resources or testing different configurations. Resources that aren't required by any enabled metric are simply ignored by `resolve()` and therefore aren't used by the QC pipeline.

src/wf_psf/tests/test_quality_control/conftest.py
jeipollack: Yes, the factory implementation is the same across the tests and is not modified by individual tests. I left it at the default function scope because it is inexpensive to construct and the factory itself is stateless; each test still creates its own `QualityControlConfig` instance. I agree that session scope would also work, but I don't think it provides a meaningful benefit here.

src/wf_psf/quality_control/resources.py
roryclaydon1994: I'm not quite following, but I'm assuming examples will come in the next and MR, and at present, this seems like it will not cause an issue.

I think a compromise might be to add a quick debug logging call to the function to specify what was missing, what was resolved, and was provided but not required.

jeipollack: okay, no p. I can add a debug logging call and will ping you when it's been committed

jeipollack: pushed the new commit. No new test was added because the unused-resource scenario is already covered by the existing parameterized test.

PR #248, opened 2026-08-24
Generalise rejection policy configuration

### LINKED ISSUE #247: [ARCHITECTURE] Generalise rejection policy configuration
## Context

An architectural issue was identified during development of the quality
control pipeline orchestration involving the instantiation of rejection
policies from the rejection policy registry.

The current design supports only one policy type: `threshold`. The
`RejectionPolicyConfig` dataclass therefore defines `threshold` as a
configuration attribute. This limits the configuration model as additional
policy types (e.g. quantile-based policies) may require different parameters.

Generalise rejection policy configuration to support registry-selected policy
implementations and policy-specific parameters.

## Scope

Tasks are:

- Add a policy identifier to `RejectionPolicyConfig`.
- Replace the hard-coded `threshold` field with a generic `params` mapping.
- Update `parse_rejection_policy_config()`.
- Update the YAML configuration examples and fixtures.
- Update rejection-policy tests.
- Update rejection policy instantiation to obtain the implementation from the
  configured policy identifier (deferred to the pipeline orchestration PR #243)

jeipollack:
## Summary

This PR generalises the rejection policy configuration section by replacing the hard-coded threshold field with a generic policy mapping.  The current hard-coded representation limits the configuration model as additional policy types (e.g. quantile-based policies) may require different parameters.

Closes #247 

## What’s changed

- Replace the hard-coded threshold field with a generic policy mapping
- Update `RejectionPolicyConfig` to support policy-specific parameters
- Update `parse_rejection_policy_config()`
- Update the YAML configuration examples and fixtures
- Update rejection-policy tests
- Use backticks in error messages to highlight configuration parameter names

## How to test / verify

- Check CI tests pass.

## Scope

> Indicate the type of PR:

- [ ] Feature  
- [ ] Bug fix  
- [ ] Hotfix  
- [X] Internal / refactor  
- [ ] Release

## Changelog

No changelog fragment is added because the final user-facing configuration structure is still being developed. The changelog entry will be added with the PR that finalises the configuration structure.

## Reviewer Checklist

- [x] CI tests are passing

Continue development of the pipeline orchestration PR, incorporating the generalised rejection policy configuration into rejection policy instantiation.

### DIFF
--- a/src/wf_psf/quality_control/config.py
+++ b/src/wf_psf/quality_control/config.py
@@ -47,12 +47,12 @@ class RejectionPolicyConfig:
     enabled : bool
         Whether rejection policy is enabled.
 
-    threshold : float | None
-        Numeric threshold used by the rejection policy, or None if not applicable.
+    policy : dict[str, Any]
+        Rejection policy configuration keyed by policy type.
     """
 
     enabled: bool = False
-    threshold: float | None = None
+    policy: dict[str, Any] = field(default_factory=dict)
 
 
 @dataclass
@@ -153,7 +153,7 @@ def parse_metrics_config(
         enabled = cfg.get("enabled", True)
 
         if not isinstance(enabled, bool):
-            raise TypeError(f"Metric enabled flag for '{name}' must be boolean.")
+            raise TypeError(f"Metric `enabled` flag for '{name}' must be boolean.")
 
         params = cfg.get("params", {})
 
@@ -198,6 +198,9 @@ def parse_rejection_policy_config(
     TypeError
         If the configuration section or a policy configuration is not a
         mapping.
+
+    ValueError
+        If the policy configuration for a metric does not specify exactly one policy.
     """
     if section is None:
         return {}
@@ -207,13 +210,51 @@ def parse_rejection_policy_config(
 
     policies = {}
 
-    for name, cfg in section.items():
+    for metric_name, cfg in section.items():
         if not isinstance(cfg, Mapping):
             raise TypeError(
-                f"Rejection policy configuration '{name}' must be a mapping."
+                f"Rejection policy configuration '{metric_name}' must be a mapping."
             )
 
-        policies[name] = RejectionPolicyConfig(**cfg)
+        enabled = cfg.get("enabled", True)
+
+        if not isinstance(enabled, bool):
+            raise TypeError(
+                f"Rejection policy `enabled` flag for '{metric_name}' must be boolean."
+            )
+
+        if not enabled:
+            policies[metric_name] = RejectionPolicyConfig(enabled=False)
+            continue
+
+        policy = cfg.get("policy", {})
+
+        if not isinstance(policy, Mapping):
+            raise TypeError(
+                f"Rejection policy `policy` field for '{metric_name}' must be a mapping."
+            )
+
+        if len(policy) != 1:
+            raise ValueError(
+                f"Rejection policy configuration for '{metric_name}' "
+                "must specify exactly one policy type."
+            )
+
+        policy_name, policy_params = next(iter(policy.items()))
+
+        if not isinstance(policy_name, str):
+            raise TypeError(
+                f"Rejection policy identifier '{policy_name}' for '{metric_name}' must be a string."
+            )
+
+        if not isinstance(policy_params, Mapping):
+            raise TypeError(
+                f"Rejection policy parameters for '{metric_name}' must be a mapping."
+            )
+
+        policies[metric_name] = RejectionPolicyConfig(
+            enabled=enabled, policy=dict(policy)
+        )
 
     return policies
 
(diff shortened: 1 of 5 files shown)


src/wf_psf/quality_control/config.py
```diff
@@ -153,7 +153,7 @@ def parse_metrics_config(
         enabled = cfg.get("enabled", True)
```
roryclaydon1994: Just noticed all metrics are `True` by default? Just wondering if `False`  would match a user's expectation more closely?

src/wf_psf/quality_control/config.py
roryclaydon1994: We could change `name` here to be more informative like the `metrics_name` added in the `parse_rejection_policy_config` function

src/wf_psf/quality_control/config.py
```diff
@@ -207,13 +210,51 @@ def parse_rejection_policy_config(
...
 
-    for name, cfg in section.items():
+    for metric_name, cfg in section.items():
         if not isinstance(cfg, Mapping):
             raise TypeError(
-                f"Rejection policy configuration '{name}' must be a mapping."
+                f"Rejection policy configuration '{metric_name}' must be a mapping."
+            )
+
+        enabled = cfg.get("enabled", True)
```
roryclaydon1994: Same q as above

src/wf_psf/quality_control/config.py
```diff
@@ -207,13 +210,51 @@ def parse_rejection_policy_config(
...
+            raise TypeError(
+                f"Rejection policy `policy` field for '{metric_name}' must be a mapping."
             )
 
-        policies[name] = RejectionPolicyConfig(**cfg)
+        if len(policy) != 1:
+            raise ValueError(
+                f"Rejection policy configuration for '{metric_name}' "
+                "must specify exactly one policy type."
+            )
```
roryclaydon1994, line 251: Does the default always fail?

```
Python 3.13.14 (main, Jun 23 2026, 15:18:27) [Clang 22.1.3 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> len({})
0
```

roryclaydon1994: Is the restriction to one policy per metric name for simplicity?

roryclaydon1994 (commented):
## Review Feedback

Changes look good, consistently good documentation throughout. Design is flexible and the type enforcement is likely to limit the error space.

In terms of design, I'm wondering if the constraint of one policy per metric could be relaxed such that one pass through the data could handle multiple outputs, one per policy? Also, given that the typing is strong, perhaps reading the configs to a dataclass first to handle the type handling would give a consistent API to the configs and would allow the parsing functions to focus on the policy extraction?

jeipollack:
Thanks, @roryclaydon1994 for the review!

I agree that having the configuration represented as typed dataclasses as early as possible would provide a cleaner API. My current understanding is that this would require either a typed representation of the full YAML configuration before section parsing, or moving the mapping-to-dataclass conversion into the individual configuration classes. I'd like to explore that separately rather than introduce that architectural change into this PR, since the current parsers provide the boundary between the raw YAML mappings and the validated `QualityControlConfig`.

For multiple policies per metric, I think enforcing one policy for now keeps the configuration and execution model simple. If a concrete use case emerges for multiple policies, we can revisit the configuration and combine the policy outputs at that point.

I will now address each individual conversation thread :-)

Done! Ready for you to have a second pass

src/wf_psf/quality_control/config.py
jeipollack: Good point! Configuration is opt-in to running the pipeline rather than opt-out.

src/wf_psf/quality_control/config.py
jeipollack: Agreed. I used name here because the surrounding metrics mapping already provides the context, but `metric_name` makes the identifier's role clearer and is consistent with the rejection-policy parser. I'll update this.

src/wf_psf/quality_control/config.py
jeipollack: yeah, for now it is. We can extend to multiple policies when the need arises.

src/wf_psf/quality_control/config.py
jeipollack: updated to False

src/wf_psf/quality_control/config.py
roryclaydon1994: Please check the default always fail [above](https://github.com/CosmoStat/wf-psf/pull/248#discussion_r3853004885)

jeipollack: Just updated the unit test to include null case.

jeipollack: Cleared up misunderstanding in Slack, I removed the empty dict default and instead added a new check to verify policy field exists in configuration which raises a `ValueError` if missing, added a unit test for that, a unit test to check that `policy` is a Mapping, and I removed a unit test that already tested the `policy: {}` as it was merged into the existing "must specify one policy" test.

PR #250, opened 2026-08-31
Introduce resource preparation

### LINKED ISSUE #249: [REFACTOR] Resource preparation
## Description of Refactoring/Improvement

This task extends the `Resources` class, introduced in #244, to prepare resources required by enabled quality control metrics when they are not supplied by the pipeline caller.

The current `Resources.resolve()` method identifies missing resources but raises `NotImplementedError` because resource preparation is not yet implemented. This task will implement the preparation step and integrate it into resource resolution.

## Scope

* Add a `prepare_resources()` helper to the `Resources` class.
* Define how missing resource identifiers are mapped to their corresponding preparation mechanism.
* Prepare missing resources from the validated quality control configuration.
* Update `Resources.resolve()` to use `prepare_resources()` for resources not supplied by the caller.
* Preserve caller-provided resources without unnecessary preparation.
* Return a complete mapping of resources required by enabled quality metrics.
* Update unit tests for resource preparation and resolution.
* Update integration tests as appropriate.
* Update documentation and configuration examples if the preparation interface introduces user-visible configuration.

### Separation of resolution and preparation

`Resources.resolve()` should remain the public resource-management API. It should determine which resources are required, identify resources already supplied by the caller, and delegate preparation of missing resources to `prepare_resources()`.

Conceptually:

```text
required resources
       │
       ├── provided ───────────────→ resolved
       │
       └── missing ──→ prepare ────→ resolved
```

### Configuration-driven preparation

Resource preparation should be based on the validated `QualityControlConfig` rather than requiring the caller to know how individual resources are constructed.

The resource configuration should therefore provide sufficient information to identify and prepare each required resource.

### Caller-provided resources

Caller-provided resources take precedence over prepared resources.

`Resources.resolve()` first compares the resources required by enabled quality metrics with those supplied by the pipeline caller. Only resources identified as missing are passed to `prepare_resources()` for preparation.

This ensures that an already-supplied resource is never unnecessarily regenerated or replaced by a prepared resource.

### Resource-specific preparation

Preparation mechanisms may differ between resource types. The design should avoid coupling `Resources` to the implementation details of every possible resource where practical, so that additional resource types can be introduced without substantially changing the resolution interface.

### Missing and unavailable resources

The implementation should distinguish between:

* resources supplied by the caller;
* resources that are missing and can be prepared from configuration;
* resources that are required but cannot be prepared or otherwise resolved.

If an enabled metric requires a resource that cannot be resolved, the resource-resolution stage should raise a clear error identifying the unavailable resource(s), rather than silently skipping the metric.

Skipping an enabled metric because one of its required resources is unavailable is outside the scope of this task.

### Validation

The implementation should continue to rely on the validated `QualityControlConfig` established by `QualityControlConfigHandler`. Resource preparation should not duplicate cross-section configuration validation already performed during configuration loading.

## Out of Scope

* Determining whether the quality control pipeline should be run.
* Determining how the quality control pipeline is invoked by the CLI or workflow.
* Implementing individual quality metrics or rejection policies.
* Changing the quality control configuration model unless required to support resource preparation.

## Acceptance Criteria

* `Resources.resolve()` can resolve all resources required by enabled metrics when they are either supplied by the caller or can be prepared from configuration.
* Caller-provided resources are reused rather than unnecessarily prepared.
* Missing resources are prepared through the resource preparation mechanism.
* Unresolvable required resources produce a clear error.
* Existing resource-resolution behaviour remains covered by tests.
* New resource-preparation behaviour is covered by unit tests.
* Relevant integration tests pass.

jeipollack:
## Summary

Introduce resource preparation as part of resource resolution, allowing missing required resources to be prepared through registered resource preparers. 

Closes #249 

## What’s changed

- Add `RESOURCE_PREPARERS` registry for resource preparation callables
- Add `prepare_resources` helper to invoke resource preparers with resource configuration and dataset
- Extend `resolve()` to prepare missing resources and combine them with caller-provided resources
- Add/update unit tests for resource preparation and resource resolution
- Extend `RESOURCES_DEFAULT` in `conftest.py` to support resource preparation tests

## How to test / verify

- Run tests in `resources_test.py`
- CI passes.

## Scope

> Indicate the type of PR:

- [ ] Feature  
- [ ] Bug fix  
- [ ] Hotfix  
- [X] Internal / refactor  
- [ ] Release

## Changelog

## Reviewer Checklist

- [x] CI tests are passing

The resource-preparation mechanism currently provides the orchestration layer through `RESOURCE_PREPARERS`, but concrete resource preparers still need to be implemented.

In particular, a callable for preparing `psf_models` (e.g. through the PSF inference pipeline) will be added in a follow-up PR once the appropriate inference API has been defined.

### DIFF
--- a/src/wf_psf/quality_control/resources.py
+++ b/src/wf_psf/quality_control/resources.py
@@ -8,13 +8,16 @@ and resolving those requirements against resources supplied by the caller.
 """
 
 from __future__ import annotations
-from collections.abc import Mapping
+from collections.abc import Mapping, Callable
 from typing import Any
 from wf_psf.quality_control.config import QualityControlConfig
 
 import logging
 
 logger = logging.getLogger(__name__)
+
+ResourcePreparer = Callable[[Any, dict[str, Any]], Any]
+RESOURCE_PREPARERS: dict[str, ResourcePreparer] = {}
 
 
 class Resources:
@@ -46,9 +49,41 @@ class Resources:
             for resource in metric.required_resources
         }
 
+    def prepare_resources(self, missing: set, dataset: Any | None) -> dict[str, Any]:
+        """Prepare resources required by enabled quality metrics.
+
+        Parameters
+        ----------
+        missing: set
+            Unique resource identifiers required by enabled quality metrics.
+
+        dataset: Any
+            Dataset or data container required to prepare resources.
+
+        Returns
+        -------
+        dict[str, Any]
+            Prepared resources indexed by resource identifier.
+        """
+        resources_config = self.config.resources.available
+
+        prepared_resources = {}
+        for identifier in missing:
+            family, variant = identifier.split(".", 1)
+
+            resource_config = resources_config[family][variant]
+
+            preparer = RESOURCE_PREPARERS[family]
+            prepared_resource = preparer(dataset, resource_config)
+
+            prepared_resources[identifier] = prepared_resource
+
+        return prepared_resources
+
     def resolve(
         self,
         provided: Mapping[str, Any] | None = None,
+        dataset: Any | None = None,
     ) -> dict[str, Any]:
         """Resolve resources required by enabled quality metrics.
 
@@ -57,17 +92,15 @@ class Resources:
         provided : Mapping[str, Any] or None
             Ready-to-use resources supplied by the pipeline caller.
 
+        dataset : Any or None
+            Dataset or data container required to prepare missing resources.
+
         Returns
         -------
         dict[str, Any]
             Resources required by enabled quality metrics and supplied by the
             caller.
 
-        Raises
-        ------
-        NotImplementedError
-            If required resources are not supplied by the caller. Preparation of
-            missing resources is not yet implemented.
         """
         required = self.get_required()
         provided = {} if provided is None else provided
@@ -77,8 +110,13 @@ class Resources:
             for resource in required
             if resource in provided
         }
+
         missing = required - provided.keys()
         unused = provided.keys() - required
+
+        if missing:
+            prepared_resources = self.prepare_resources(missing, dataset)
+            resolved.update(prepared_resources)
 
         logger.debug(
             "Resource resolution: resolved=%s, missing=%s, unused=%s",
@@ -86,10 +124,5 @@ class Resources:
             sorted(missing),
             sorted(unused),
         )
-
-        if missing:
-            raise NotImplementedError(
-                f"Required resources are not available: {sorted(missing)}"
-            )
 
         return resolved
--- a/src/wf_psf/tests/test_quality_control/conftest.py
+++ b/src/wf_psf/tests/test_quality_control/conftest.py
@@ -29,7 +29,8 @@ def qc_config_factory():
                 "psf_models": {
                     "standard": {
                         "inference_config": "inference_standard.yaml",
-                    }
+                    },
+                    "oversampled": {"inference_config": "inference_oversampled.yaml"},
                 }
             }
         )
--- a/src/wf_psf/tests/test_quality_control/resources_test.py
+++ b/src/wf_psf/tests/test_quality_control/resources_test.py
@@ -6,9 +6,12 @@ This module contains unit tests for the quality control resources module.
 
 """
 
+import numpy as np
 import pytest
+from unittest.mock import Mock, call
+
 from wf_psf.quality_control.config import QualityMetricConfig
-from wf_psf.quality_control.resources import Resources
+from wf_psf.quality_control.resources import RESOURCE_PREPARERS, Resources
 
 
 def test_get_required(qc_config_factory):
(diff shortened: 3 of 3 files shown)


src/wf_psf/quality_control/resources.py
```diff
@@ -46,9 +49,41 @@ def get_required(self) -> set[str]:
             for resource in metric.required_resources
         }
 
+    def prepare_resources(self, missing: set, dataset: Any | None) -> dict[str, Any]:
```
roryclaydon1994: For the typehint of dataset, I assume None is a subtype of Any?

src/wf_psf/quality_control/resources.py
```diff
@@ -46,9 +49,41 @@ def get_required(self) -> set[str]:
...
+        dict[str, Any]
+            Prepared resources indexed by resource identifier.
+        """
+        resources_config = self.config.resources.available
+
+        prepared_resources = {}
+        for identifier in missing:
+            family, variant = identifier.split(".", 1)
+
+            resource_config = resources_config[family][variant]
```
roryclaydon1994: Would this operation happen a lot? I'm just wondering if a small dataclass with a family and variant attribute would help, or if the resources config might be able to be equipped with a get method that handles the splitting and access?

src/wf_psf/quality_control/resources.py
```diff
@@ -8,14 +8,17 @@
...
+from collections.abc import Mapping, Callable
 from typing import Any
 from wf_psf.quality_control.config import QualityControlConfig
 
 import logging
 
 logger = logging.getLogger(__name__)
 
+ResourcePreparer = Callable[[Any, dict[str, Any]], Any]
+RESOURCE_PREPARERS: dict[str, ResourcePreparer] = {}
```
roryclaydon1994, line 21: It looks like this is another registry, so I'm guessing at some point there will be a build_resources_preparer_registry added to match the other ones?

src/wf_psf/quality_control/resources.py
```diff
@@ -46,9 +49,41 @@ def get_required(self) -> set[str]:
...
+        """
+        resources_config = self.config.resources.available
+
+        prepared_resources = {}
+        for identifier in missing:
+            family, variant = identifier.split(".", 1)
+
+            resource_config = resources_config[family][variant]
+
+            preparer = RESOURCE_PREPARERS[family]
```
roryclaydon1994: I would suggest adding some guards or try/excepts here as at the moment this is always used uninitialised. 

Maybe one for it being completely empty. The key error is maybe sufficient on its own, but catching it and making the error more specific could be helpful e.g. how to add a preparer

roryclaydon1994 (changes requested):
## Review Feedback

Overall design fits with the modular approach and is sufficiently flexible to achieve adding arbitrary resource preparers.

My main question is how the resource preparer registry will be populated? Is this something that is built with most normal preparers and then the user can add custom ones?

At present, I think an uninitialised resource preparer will always be requested, so I think we should think about that before accepting the PR, perhaps with something like raising an explicit error or logging a warning and returning a dummy output?

src/wf_psf/quality_control/resources.py
jeipollack: Thanks for noting this. I agree that `| None` is redundant here; I hadn't clearly thought about this relationship. I changed this to just `Any`.

I used `Any` because `RESOURCE_PREPARERS` is an extensible registry, enabling users to register their own resource preparers with their own expected dataset types. Therefore, I think `prepare_resources()` should remain agnostic about the concrete type of `dataset`.

I've also updated `missing` to `set[str]` to tighten the type specification.

src/wf_psf/quality_control/resources.py
jeipollack: The `family.variant` identifier is intentional because it provides a compact, canonical identifier for a resource requirement. The metric configuration uses this identifier, while the resources configuration is nested by family and variant.

I took some time to consider your suggestion of a small `ResourceRequirement` dataclass to represent the family/variant pair, but in this case it would mainly wrap the existing `split()` without adding much behaviour. I also considered modifying the nested `ResourcesConfig`, but it doesn't naturally own resolution of a `family.variant` identifier because it represents the configuration structure rather than the resource requirement.

Since this is currently the only place where the identifier needs to be split, I think keeping the lookup here is the simpler approach. If resource resolution becomes repeated in other parts of the pipeline, I think that would be a good point to introduce a dedicated resolver.

roryclaydon1994: > it would mainly wrap the existing split() without adding much behaviour.

That's not the goal. Any logic, or anything else that needs to be checked for such cases only needs to be done in one place. If that is a class, any related logic to identifier validation could also be moved there. For future maintainers, it's also a self describing object, so easier to understand at a glance and wouldn't always need to be split before use. This would also easily allow changes to id structure if desired at some point.

>  I also considered modifying the nested ResourcesConfig, but it doesn't naturally own resolution of a family.variant identifier because it represents the configuration structure rather than the resource requirement.

It doesn't have to own anything other than the config. The point here similar to above in that the splitting logic lives in a single, sensible place and access is controlled through there and could easily handle other dictionary/config layouts if anything changes in future.

Some examples
```
# somewhere else in code we do something like checking for incorrect ids, which now lives in the dataclass
cid = Identity(identity_from_config)  # checks for validity in __post_init__
.....
resource_config = resources_config[cid.family][cid.variant]
# or
resource_config = resources_config.from_id(cid)
```
or sticking to string
```
resource_config = resources_config.get(identifier)  # add a getter to config (possible need to inherit a dict)
```

Let me know what you think.

src/wf_psf/quality_control/resources.py
jeipollack: Yes, this is another registry, but for this one I introduced a `register_resource_preparer` decorator rather than a separate `build_resource_preparer_registry` function. The decorator allows both built-in and custom preparers to register themselves using the same mechanism.

I think this is a useful pattern for resource preparers because the registry is intended to be extensible by users of the quality-control API as well as populated with built-in preparers.

I know this is a departure from the existing registry pattern, so we could consider adopting the decorator-based approach for the existing registries in a follow-up PR.

src/wf_psf/quality_control/resources.py
jeipollack: I added the try/except with an instruction on how to populate the registry using the `register_resource_preparer` decorator. This handles the case where the registry is empty or the requested resource family is not registered, and provides a more actionable error message.

jeipollack: oops I just realise the doc string is incomplete. will push the commit shortly

src/wf_psf/quality_control/resources.py
jeipollack: Okay, you convinced me :) to implement a `ResourceIdentifier` dataclass in a dedicated `resource_identifier.py` module. It resulted in a cleaner interface for resource parsing, validation, and preparation.

I updated `config.py` and `resources.py`, added tests for `ResourceIdentifier`, and updated the existing tests accordingly. All tests are passing.

src/wf_psf/quality_control/resources.py
jeipollack: pushed.

jeipollack:
Thanks so much, @roryclaydon1994, for your thoughtful review and insightful perspectives. I implemented changes in line with your feedback, which led to some refactoring. I’m ready for you to take a second look.

src/wf_psf/quality_control/resources.py
roryclaydon1994: Happy to reconcile this later, but just to check, did we want to avoid the decorator version due to it auto-magicking things, or are we happy with it now?

Many libraries use decorators for purposes like this, so I'm happy with either solution

jeipollack: Yes, at first I/Chat thought it would be useful to be explicit in the beginning for clarity's sake. But, I also use decorators for registering pipeline tasks and different classes PSF models.

roryclaydon1994 (approved):
## Review Feedback

Nice clean up and I think the decorator looks fine. Let's get this merged :)

PR #253, opened 2026-09-09
Separate noise estimation from sample weight calculation

### LINKED ISSUE #225: Separate noise estimation from sample weight computation
**Description of Refactoring/Improvement**
The SHE validation workflow requires access to the noise estimate associated with each observation to compute goodness-of-fit metrics for the WaveDiff model.

Currently, `train_utils.calculate_sample_weights` performs two distinct operations:
- estimate the noise level using the `utils.NoiseEstimator` class;
- convert this estimate into a scalar sample weight using a median-based reduction.

The current implementation does not expose the intermediate noise estimate independently, limiting its reuse for validation metrics such as reduced χ².

**Goals and Objectives**

- [ ]  Extract the noise estimation step into a dedicated `estimate_noise_sigma` function.
- [ ]  Update calculate_sample_weights to call `estimate_noise_sigma` internally.
- [ ]  Preserve the existing weighting behaviour (median-based reduction).
- [ ]  Maintain backwards compatibility with the current training pipeline.
- [ ]  Expose the per-observation noise estimate for use in future uncertainty models and goodness-of-fit calculations.

**Current Code Behaviour**
`calculate_sample_weights` currently combines two responsibilities:
- Estimate the noise level for each observation using `utils.NoiseEstimator`, which uses a robust MAD-based estimator on unmasked background pixels.
- Convert this estimate into a scalar sample weight used during training.

The coupling of these operations prevents the noise estimate from being accessed independently by downstream validation workflows.

**Proposed Changes**

Introduce a dedicated helper:

```
estimate_noise_sigma(...)
```

which returns the per-observation noise standard deviation estimated by `utils.NoiseEstimator`.

Refactor:
```
calculate_sample_weights(...)
```

to use:
```
sigma = estimate_noise_sigma(...)
weights = ...
```
while preserving the current training behaviour.

**Expected Benefits**
- Separates data-derived noise estimation from training-specific weighting.
- Makes the uncertainty information available for future goodness-of-fit calculations.
- Improves clarity around the statistical assumptions used by the training pipeline.
- Avoids duplicating noise estimation logic in validation workflows.

**Dependencies**
- Compatibility with the SHE validation pipeline.
- Conversion from noise estimates to an inverse covariance representation remains dependent on the uncertainty model adopted by the validation workflow and is outside the scope of this refactor.

ChaitanyaChawak:
## Summary

- Separates noise estimation from sample-weight computation in train_utils so the per-observation noise sigma can be reused by the SHE validation workflow.
- Closes #225 

## What’s changed

- Added `train_utils.estimate_noise_sigma` method which returns the per-observation noise standard deviation.
- Refactored `calculate_sample_weights` to call `estimate_noise_sigma` internally.
- Added private `_is_masked_loss` helper to help simplify `calculate_sample_weights` method.
- Added tests covering the new methods

## How to test / verify

- Check if CI passes

## Scope

> Indicate the type of PR:

- [ ] Feature  
- [ ] Bug fix  
- [ ] Hotfix  
- [x] Internal / refactor  
- [ ] Release

> This PR is part of the "Compatibility with Euclid SHE Validation" milestone.

## Changelog

## Reviewer Checklist

- [ ] The PR targets the correct base branch (`develop`, or `main` for release PRs)
- [ ] The PR is assigned to the developer
- [ ] Appropriate labels are applied
- [ ] The PR is included in relevant projects and/or milestones
- [ ] Description clearly explains what has changed
- [ ] Issue references included, if applicable
- [ ] Code and documentation adhere to current standards (`ruff`)
- [ ] Documentation updates included, if relevant
- [ ] CI tests are passing
- [ ] All reviewer comments have been addressed

> Will use this abstracted out `estimate_noise_sigma` method for in the she_validation_adaptor required by the SHE Validation pipeline


src/wf_psf/training/train_utils.py
```diff
@@ -367,6 +367,81 @@ def configure_optimizer_and_loss(
...
+        masks = np.array(1 - outputs[..., 1], dtype=bool)
+        imgs_std = np.array(
+            [std_est.estimate_noise(_im, _win) for _im, _win in zip(images, masks)]
+        )
+    else:
+        logger.info("Estimating noise standard deviation for images..")
+        # Estimate noise standard deviation
+        imgs_std = np.array([std_est.estimate_noise(_im) for _im in outputs])
+
+    return imgs_std
```
jeipollack, line 442: I think this extraction exposes a separation-of-responsibilities issue worth considering.

`estimate_noise_sigma()` currently takes `outputs` and `loss`, and `_is_masked_loss()` is used to determine whether the outputs contain `[image, mask]`. This means the noise-estimation helper is coupled to the training-specific representation of the data and to the particular loss function being used.

Could we instead have the training code resolve the image/mask data from `outputs`, and have the reusable noise-estimation functionality operate on images with an optional mask? In fact, since `NoiseEstimator` already encapsulates the exclusion window and the MAD-based noise estimation, I wonder whether the batch noise-estimation functionality belongs as a method of `NoiseEstimator` itself, perhaps named `estimate_noise_batch()` or whatever you think is most suitable.

Given that `NoiseEstimator` is general image/noise-analysis functionality, I also wonder if it would make sense to move the class out of `utils/utils.py` into something like `utils/noise.py`.

This would allow the same functionality to be used directly by the SHE validation workflow without introducing a dependency on `train_utils.py` or the training-specific loss representation.

jeipollack (changes requested):
Thanks, Chai, for your work on this PR. You implemented all the right ingredients to arrive at the correct outcome. There is just a question of the reusability of `estimate_noise_sigma()` as a general image-analysis tool, since it is currently coupled to the training logic.

One general point I'd like to ask you to keep in mind for this and future tasks: when the implementation exposes something that doesn't fit the conceptual responsibility of the code you're adding, stop and consider whether the design should change. Then make an intentional decision about whether that change belongs in the current scope.

If the improvement is small and directly related to the task, it can be addressed in the current PR. If it's larger or more loosely related, it can be captured as a follow-up issue. If it's necessary to implement the task cleanly, then the refactoring should be done as part of the work, even if that means separating it into a prerequisite PR.

This is the strategy I've been applying in the Quality Control Pipeline Milestone, and I think it has been working well.

src/wf_psf/training/train_utils.py
ChaitanyaChawak: thanks @jeipollack for this thorough review! I have updated the PR based on your feedback :smile:

ChaitanyaChawak:
Thanks a lot @jeipollack for this thorough review! As suggested, I have moved all the NoiseEstimator utilities into a noise.py file, and added a separate estimate_noise_batch method. This method is now no longer depends on the loss or any training-specific elements. Thanks a lot for pointing this shortcoming in the original PR. I have also updated the tests for the updated parts, and the PR is now ready for a second review pass :pray:

jeipollack:
Thanks @ChaitanyaChawak for the updates. Moving `NoiseEstimator` into `noise.py` and adding the batch estimation method is a good improvement. I think there are still a few places where we can simplify the separation of responsibilities further.  I'm also getting better at spotting these things now than when I first began working on the project. I admit that I am partially responsible for these code bits, which contained mixed logic. 🙈 

To start, I think the training-specific resolution of `outputs` should happen in `general_train_cycle()`, immediately before calling `calculate_sample_weights()`. The loss is only needed here to determine how the training `outputs` representation should be interpreted (i.e. whether to extract an image and mask). Once that has been resolved, `calculate_sample_weights()` doesn't need to know about the loss function at all.

This would give `calculate_sample_weights()` a much simpler contract along the lines of:

```python
calculate_sample_weights(
    images,
    masks=None,
    apply_sigmoid=False,
    sigmoid_max_val=5.0,
    sigmoid_power_k=1.0,
)
```

I also think `use_sample_weights` is better handled by the caller. Rather than having `calculate_sample_weights()` decide whether it should calculate anything, `general_train_cycle()` could conditionally call it and otherwise leave `sample_weight = None`. That would give the function a single responsibility: calculating the weights from the supplied images/masks.

Finally, I wonder whether the construction of `NoiseEstimator` could encapsulate more of the noise-estimation configuration. At the moment the caller needs to know both how to extract the image dimensions and how the exclusion-window radius is calculated:

```python
img_dim = (outputs.shape[1], outputs.shape[2])
win_rad = np.ceil(outputs.shape[1] / 3.33)
std_est = NoiseEstimator(img_dim=img_dim, win_rad=win_rad)
```

The window-radius calculation in particular feels like knowledge that belongs with `NoiseEstimator`, rather than with the training code. It may be worth considering whether the estimator can derive this from the image dimensions itself, while retaining the ability to configure the radius if needed.

Putting this all together, I think this would leave us with a clearer boundary: `general_train_cycle()` understands the training representation and loss; `calculate_sample_weights()` understands how to turn image noise into training weights; and `NoiseEstimator` owns the actual noise-estimation logic and its configuration.  With this tidy-up, we may not even need `estimate_noise_sigma` method because the functionality is already handled by `estimate_noise_per_batch`, which I think you can re-use in your SHE Validation adapter.

Hope that makes sense! And, apologies for the back and forth. Just know I have such exchanges with Rory, too!  ;-)

ChaitanyaChawak:
yess, I think that makes a lot of sense! I've updated the `calculate_sample_weights()` function to make it much more simple and the training-specific resolution of the `output` is now done in `general_train_cycle` just before the sample weights are calculated.

I have also moved the window radius calculation in `NoiseEstimator` by introducing an optional win_rad parameter during init. The default value of this parameter is calculated in a newly introduced `default_win_rad` method using the provided `img_dim`.

Finally, as you correctly predicted, all these changes rendered `estimate_noise_sigma` obsolete and it has now been removed.

I've also updated the relevant sections of the `train_utils_test.py` to correctly test for the changes made in this PR and fix the CI pytest failure :v:

src/wf_psf/training/train_utils.py
```diff
@@ -722,15 +726,26 @@ def general_train_cycle(
...
+    if use_sample_weights:
+        sample_weight = calculate_sample_weights(
+            images,
+            masks,
+            apply_sigmoid,
+            sigmoid_max_val,
+            sigmoid_power_k,
+        )
+    else:
+        sample_weight = None
```
jeipollack: One small thought on this part: since we only need to resolve the training-specific `outputs` representation when `use_sample_weights` is `True`, I think the resolution could happen inside that conditional:

```python
if use_sample_weights:
    images, masks = _resolve_training_outputs(outputs, loss)
    sample_weight = calculate_sample_weights(
        images,
        masks,
        apply_sigmoid,
        sigmoid_max_val,
        sigmoid_power_k,
    )
else:
    sample_weight = None
```

I also wonder if `_resolve_training_outputs()` could encapsulate the `_is_masked_loss()` check and the extraction of `images`/`masks`. That would avoid introducing a nested conditional block, and would keep the representation-specific logic together while leaving `general_train_cycle()` focused on the higher-level training flow.

ChaitanyaChawak: done!

src/wf_psf/tests/test_training/train_utils_test.py
```diff
@@ -34,6 +34,9 @@ def mock_noise_estimator():
         mock_instance.estimate_noise.side_effect = lambda img, mask=None: np.std(
             img
         )  # Mock behavior
```
jeipollack: I don't think we need to mock `estimate_noise()` here, since `estimate_noise_batch()` is also mocked and therefore the actual implementation of `estimate_noise_batch()` (which calls `estimate_noise()`) is never reached.

I think we can just keep the `estimate_noise_batch` mock for these tests.

src/wf_psf/tests/test_training/train_utils_test.py
```diff
@@ -368,6 +361,71 @@ def test_general_train_cycle_handles_cycle_def(
...
+    before calculate_sample_weights ever sees the data."""
+    mock_model = mock_test_setup["mock_model"]
+    inputs = mock_test_setup["inputs"]
+
+    batch_size, height, width = 4, 8, 8
+    images = np.random.rand(batch_size, height, width)
+    raw_masks = np.random.randint(0, 2, size=(batch_size, height, width))
+    outputs = np.stack([images, raw_masks], axis=-1)
+
+    with patch("wf_psf.training.train_utils.calculate_sample_weights") as mock_calc:
```
jeipollack: Small naming suggestion: could we use a slightly more descriptive name here, e.g. `mock_calculate_sample_weights`.  I find this easier to read than `mock_calc`, particularly when scanning the test and assertions.

src/wf_psf/tests/test_training/train_utils_test.py
```diff
@@ -368,6 +361,71 @@ def test_general_train_cycle_handles_cycle_def(
...
+    np.testing.assert_array_equal(called_masks, np.array(1 - raw_masks, dtype=bool))
+
+
+def test_general_train_cycle_skips_sample_weights_when_disabled(mock_test_setup):
+    """general_train_cycle leaves sample_weight as None and never calls
+    calculate_sample_weights when use_sample_weights is False."""
+    mock_model = mock_test_setup["mock_model"]
+
+    with (
+        patch("wf_psf.training.train_utils.calculate_sample_weights") as mock_calc,
```
jeipollack: same here

src/wf_psf/tests/test_utils/noise_test.py
```diff
@@ -0,0 +1,156 @@
+import numpy as np
+from wf_psf.utils.noise import NoiseEstimator
+
+
+def test_initialization():
+    """Test if NoiseEstimator initializes correctly."""
+    img_dim = (50, 50)
+    win_rad = 10
+    estimator = NoiseEstimator(img_dim, win_rad)
```
jeipollack, line 9: Missing test case `NoiseEstimator(img_dim)` to see that `self.win_rad` is initialised correctly.

src/wf_psf/tests/test_training/train_utils_test.py
ChaitanyaChawak: updated!

src/wf_psf/tests/test_training/train_utils_test.py
ChaitanyaChawak: done!

src/wf_psf/tests/test_training/train_utils_test.py
ChaitanyaChawak: done

src/wf_psf/tests/test_utils/noise_test.py
ChaitanyaChawak: added all the missing test cases!

src/wf_psf/training/train_utils.py
```diff
@@ -367,31 +368,85 @@ def configure_optimizer_and_loss(
...
+        masks (``outputs[..., 1]`` inverted); otherwise ``None``.
+    """
+    if _is_masked_loss(loss):
+        images = outputs[..., 0]
+        masks = np.array(1 - outputs[..., 1], dtype=bool)
+    else:
+        images = outputs
+        masks = None
+
+    return images, masks
```
jeipollack, line 422: Don't hate the messenger (or reviewer) 🙈 , but I spotted something else:

Since `_resolve_training_outputs()` only needs to know whether the configured loss is a masked loss, do we need to pass the whole loss function here? Could we pass `loss.name` instead and have `_is_masked_loss()` operate on the name? This would make the helper's dependency on the loss representation a little clearer and avoid passing more information than it needs.

Also, rather than checking specifically for "masked_mean_squared_error", I wonder if we could use the naming convention to identify masked losses (e.g. `loss_name.startswith("masked_")`). That would allow the same output representation handling to work for other masked loss functions without needing to update this helper.

src/wf_psf/training/train_utils.py
```diff
@@ -723,14 +760,17 @@ def general_train_cycle(
...
-    sample_weight = calculate_sample_weights(
-        outputs,
-        use_sample_weights,
-        loss,
-        apply_sigmoid,
-        sigmoid_max_val,
-        sigmoid_power_k,
-    )
+    if use_sample_weights:
+        images, masks = _resolve_training_outputs(outputs, loss)
```
jeipollack: Maybe you can pass loss.name here instead, since `_resolve_training_outputs()` only needs the loss name?

src/wf_psf/training/train_utils.py
jeipollack: maybe we don't even need the `is_masked_loss()` helper anymore. it can just get resolved in the resolved method.

src/wf_psf/training/train_utils.py
jeipollack: Maybe you can pass `loss.name` here instead, since `_resolve_training_outputs()` only needs the loss name?

jeipollack:
Hi @ChaitanyaChawak, thanks for your positivity in our series of iterations. I spotted only 1-2 small code items that I think could be cleaned up. And noted now that I don't see a changelog fragment for this PR. we're also adding batch noise-estimation functionality to `NoiseEstimator`, so I think this is worth capturing in the changelog.

One these are completed I would be happy to approve this PR and merge.

src/wf_psf/training/train_utils.py
ChaitanyaChawak: thanks for catching this, that makes sense! I've now updated the workflow to reflect this :smile:

ChaitanyaChawak:
@jeipollack : Done!

changelog.d/20260909_155518_chaitanyachawak001_225_separate_noise_estimation_from_sample_weight_computation.md
```diff
@@ -0,0 +1,16 @@
...
+
+### New features
+
+- Added `NoiseEstimator.estimate_noise_batch` for batch noise estimation, which is reusable outside training (e.g. SHE validation) since it no longer depends on the training loss/output representation.
+
+### Internal changes
+
+- Moved `NoiseEstimator` from `utils.utils` to a dedicated `utils.noise` module, and initialised it with a default exclusion-window radius (`NoiseEstimator.default_win_rad`).
+- Simplified `train_utils.calculate_sample_weights` to `calculate_sample_weights(images, masks=None, ...)`. It no longer needs to know about `loss` or `use_sample_weights`. `general_train_cycle` now resolves the training output representation.
+- Removed `train_utils.estimate_noise_sigma`, which is now redundant after the refactors.
```
jeipollack: I think the CHANGELOG should be with respect to a release. Since `estimate_noise_sigma` does not exist in the release version nor was it even committed to the `develop`, its removal does not need to be listed.

src/wf_psf/tests/test_training/train_utils_test.py
```diff
@@ -183,21 +180,38 @@ def test_calculate_sample_weights_high_variance(mock_noise_estimator):
...
 
 @pytest.mark.parametrize(
-    "loss, expected",
-    [
-        (None, False),
-        ("mean_squared_error", False),
-        ("masked_mean_squared_error", True),
-        (train_utils.MaskedMeanSquaredError(), True),
-    ],
+    "loss_name", [None, "mean_squared_error", "some_other_loss"]
```
jeipollack: better if your null test included some fuzziness like "non_masked_loss"  or "unmasked_mse_"

jeipollack:
Hi @ChaitanyaChawak, I looked over your latest commits, and left a couple of comments concerning the CHANGELOG and a unit test.

changelog.d/20260909_155518_chaitanyachawak001_225_separate_noise_estimation_from_sample_weight_computation.md
ChaitanyaChawak: updated the CHANGELOG :+1:

src/wf_psf/tests/test_training/train_utils_test.py
ChaitanyaChawak: updated!

ChaitanyaChawak:
@jeipollack thanks for the speedy review! I've made the changes acc to your suggestions :raised_hands:

changelog.d/20260909_155518_chaitanyachawak001_225_separate_noise_estimation_from_sample_weight_computation.md
jeipollack: Looking at your latest changes, I think it's worth mentioning the `masked_` naming convention here. The resolution of the training output representation now depends on whether `loss.name` starts with `masked_`; as your test shows, `masked` is not accepted. So this is an important part of the contract for anyone adding or modifying a loss function.

ChaitanyaChawak: updated the CHANGELOG to mention the `masked_` naming convention

Issue #254, opened 2026-09-15
[TASK] Design and implement pixel mask quality metrics


jeipollack:
## Parent Issue

#223 — [NEW FEATURE] Star core pixel masking outlier rejection

## Problem

The pixel-mask outlier rejection feature requires quantitative metrics describing the extent and location of masked pixels within each stellar image.

A single total masked-pixel fraction is not sufficient to identify problematic observations, since masking in the peripheral regions of an image is less concerning than masking pixels within the stellar core.

The metric design should therefore distinguish between overall mask coverage and masking within a configurable region around the stellar source.

## Proposed Solution

Introduce a `PixelMaskMetrics` component responsible solely for calculating quality metrics from pixel masks.

The initial metric set should include:

* **Total masked pixels** — number of masked pixels in the image.
* **Total masked fraction** — fraction of image pixels that are masked.
* **Core masked pixels** — number of masked pixels within a configurable aperture around the source.
* **Core masked fraction** — fraction of pixels within the aperture that are masked.

The core aperture should be configurable in terms of:

* centre, e.g. source centroid or image centre;
* aperture size, expressed relative to a configurable PSF scale (e.g. a radius in units of σ).

The metric implementation should calculate these quantities without making any decision about whether an observation is an outlier.

Thresholding and classification should be handled separately by the outlier-rejection framework.

## Design Considerations

The metric API should:

* operate on the existing pixel-mask representation used by the PSF preprocessing pipeline;
* support batched observations;
* return metrics aligned with the input observations;
* allow additional pixel-mask metrics to be added without modifying the rejection framework;
* keep metric calculation separate from thresholding and rejection decisions.

The terminology should distinguish the metric computation from the later concept of "mask obscuration" or "outlier rejection". `PixelMaskMetrics` is therefore proposed as the metric component name.

## Scope

This issue covers:

* [ ] Define the `PixelMaskMetrics` API
* [ ] Define the initial metric set
* [ ] Define the configurable core aperture
* [ ] Implement the metric calculations
* [ ] Add unit tests
* [ ] Document the metric API and behaviour

This issue does **not** cover:

* configurable rejection thresholds;
* classification of observations as outliers;
* removal of rejected observations from training/validation datasets;
* rejection reporting;
* integration into the complete outlier-rejection pipeline.

These will be addressed by the parent issue and/or follow-up tasks.

## Acceptance Criteria

* `PixelMaskMetrics` computes the agreed pixel-mask metrics for individual and batched observations.
[...]

PR #258, opened 2026-09-16
Add constructor to QualityMetric ABC class

### LINKED ISSUE #257: Add constructor to `QualityMetric` ABC class
The design of the first metric for computing pixel-mask diagnostics revealed a missing step in the quality control pipeline: metric configuration parameters are not currently passed to the initialised `QualityMetric` implementations.

This issue proposes updating the `QualityMetric` ABC and metric initialisation to allow metric implementations to receive their configured parameters. We also identified an opportunity to add type hints to clarify the interface and contract that concrete metric implementations are expected to follow.

The requested changes are:

- Add a constructor to the `QualityMetric` ABC accepting params
- Update `QualityControlPipeline._initialise_metrics()` to pass the configured metric parameters to metric constructors
- Add type hints to the `QualityMetric` ABC and its concrete implementations
- Define the `compute()` return type as `dict[str, np.ndarray]`, allowing a metric to provide multiple named diagnostics that can be evaluated by rejection policies
- Update the associated tests and fixtures

jeipollack:
## Summary

This PR adds a missing step in the quality control pipeline: passing metric configuration parameters to quality metric implementations.

Closes #257

## What’s changed

* Add a constructor to `QualityMetric` to initialise the metric parameters attribute
* Update `QualityControlPipeline._initialise_metrics()` to pass metric configuration parameters to metric class constructors
* Add type hints to `QualityMetric` and its concrete implementations
* Replace wrong object `QualityMetric.compute(dataset)` with `QualityMetric.compute(context)`
* Define the return type of `QualityMetric.compute()` as `dict[str, np.ndarray]` to support multiple named diagnostic results from a metric

## How to test / verify

* No tests were updated.
* CI passes.

## Scope

> Indicate the type of PR:

* [ ] Feature
* [ ] Bug fix
* [ ] Hotfix
* [ ] Documentation / process change
* [x] Internal / refactor
* [ ] Release

## Changelog

* [ ] Changelog fragment added (if applicable)

## Reviewer Checklist

* [x] The PR targets the correct base branch (`develop`, or `main` for release PRs)
* [x] The PR is assigned to the developer
* [x] Appropriate labels are applied
* [x] The PR is included in relevant projects and/or milestones
* [x] Description clearly explains what has changed
* [x] Issue references included, if applicable
* [x] Code and documentation adhere to current standards (`ruff`)
* [x] Documentation updates included, if relevant
* [x] CI tests are passing
* [x] All reviewer comments have been addressed

The design of `PixelMaskMetrics` (#254) calls for four diagnostic metrics to be returned as a dictionary by `PixelMaskMetrics.compute()`. This motivated the change to the `QualityMetric.compute()` return type to support multiple named diagnostic results. A rejection policy can then select which diagnostic(s) it uses in its filtering strategy.

The next step is to proceed with the implementation of `PixelMaskMetrics`.


src/wf_psf/quality_control/metrics/base.py
```diff
@@ -24,30 +27,35 @@ class QualityMetric(ABC):
...
+        context : QualityControlContext
+            Context storing the dataset and resolved resources
+            shared across the quality control pipeline.
 
         Returns
         -------
-        np.ndarray
-            One metric value per dataset sample.
+        dict[str, np.ndarray]
+            Mapping of diagnostic names to arrays containing one value per
```
roryclaydon1994, line 58: So one `QualityMetric` could return multiple outputs? Is there a unified way to query what those might be in the API?

src/wf_psf/quality_control/metrics/base.py
```diff
@@ -24,30 +27,35 @@ class QualityMetric(ABC):
...
+            dataset sample. All returned arrays must be aligned with the dataset
```
roryclaydon1994: Very minor, but I think the wording should be:

ll returned arrays *will* be aligned with the dataset

as the classes will adhere to this?

src/wf_psf/quality_control/metrics/base.py
```diff
@@ -24,30 +27,35 @@ class QualityMetric(ABC):
...
-    compute(dataset)
-        Compute the quality metric for the supplied dataset.
+    compute(context)
+        Compute the quality metric for the supplied quality control context.
 
     """
 
     name: str
 
+    def __init__(self, params: dict[str, Any]):
```
roryclaydon1994, line 42: Double check if this should also be an `abstractmethod` just in case you could create a concrete instance of an abstract class.

roryclaydon1994 (commented):
## Review Feedback

PR adds to the API for the `QualityMetric` ABC to allow customisation of the metrics parameters and also collects resources and datasets into a context which helps keeps everything together. Only three small questions on this.

src/wf_psf/quality_control/metrics/base.py
jeipollack: Yes, a `QualityMetric` can return multiple diagnostics. For example,  as I told you in Slack I am working on the implementation of`PixelMaskMetric`, which returns total masked pixels/fraction as well as core masked pixels/fraction.

My current thinking is that the rejection policy will specify which diagnostic from the metric output it should operate on, e.g. `core_masked_fraction`. I was planning to work through that API when implementing the first rejection policy.  I didn't want to add a separate output-declaration mechanism to `QualityMetric` at this stage.

src/wf_psf/quality_control/metrics/base.py
jeipollack: The original wording with “must be aligned” is meant to document the contract that implementations of QualityMetric must adhere to. I think it’s stronger than describing an assumption made by callers, no?

src/wf_psf/quality_control/metrics/base.py
roryclaydon1994: I only meant adding something like
```
@abstractmethod
@property
def diagnostic_labels(self) -> list[str]:
    pass
```
So that a user could see how to check what will be returned programmatically? Or would they need to rely on the docstring for the concrete class?

src/wf_psf/quality_control/metrics/base.py
jeipollack: I checked this with Chat, and the answer was that `compute()` already makes `QualityMetric` non-instantiable, so the constructor doesn't need to be abstract. I intentionally kept it concrete because it provides the shared initialization for all metrics.

src/wf_psf/quality_control/metrics/base.py
jeipollack: Ah yes, I see what you mean. I agree that having a programmatic way to discover the available diagnostics could be useful. Could we address this in the reporting PR? I think it would be useful to first see how reporting needs to consume the metric outputs, so we can avoid introducing a separate diagnostic declaration in the ABC that potentially duplicates the keys returned by `compute()`.

src/wf_psf/quality_control/metrics/base.py
roryclaydon1994: An abstract class should not be instantiable.

Testing locally
```
>>> from abc import ABC, abstractmethod
>>> import numpy as np
>>> from typing import Any
>>> class QualityMetric(ABC):
...     def __init__(self):
...             self.hello = True
... 
>>> hello = QualityMetric()
>>>
```

No errors...

jeipollack: but you didn't include an abstract method in your example.

I tested instantiation with the actual `QualityMetric` and got

```
In [1]: from wf_psf.quality_control.metrics.base import QualityMetric

In [2]: params = {"metrics": "sth"}

In [3]: hello = QualityMetric(params)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 hello = QualityMetric(params)

TypeError: Can't instantiate abstract class QualityMetric with abstract method compute
```

src/wf_psf/quality_control/metrics/base.py
roryclaydon1994: I'm thinking as a user if I read the documentation, I would wonder why the output must adhere to a rule, rather than saying concretely output is guaranteed to adhere to the rule.

jeipollack: okay! I understand your point is the API user's perspective. So what about:  “are aligned with the dataset samples”, which more directly describes the guarantee they can rely on. I'll change it to that if you agree.

jeipollack: I pushed the commit

